PR #3, opened 2026-05-25
Verifier Tool

mfornet:
Start support for verifier


mfornet:
Merging this as is, we are going ot give another pass of love to the verifier tool soon

PR #5, opened 2026-05-27
Changed Unexpected break targeting function trap for a br

waterball01:
I changed the code so that when a br targets the implicit function level block, Wasm returns successfully. run used to trap with "Unexpected break targeting function" causing failures when running just testsuite. This code now maps Continuation.Break to a success which I assumed was the case as in Continuation.lean it is grouped with Fallthrough and Return rather than the Traps. I also added an EarlyBr.lean example, to both be a minimal example and to learn about writing code in lean4. I tested with lake build and just testsuite. I used AI to help plan the change and navigate the codebase but I reviewed all the code written, in particular the test example in lean4 which I used AI to help for syntax.

Specifically the changes were:
interpreter/Interpreter/Wasm/Semantics.lean — two Continuation.Breaks in run now produce a success.
interpreter/Interpreter/Wasm/Semantics/Lemmas.lean — run_eq was also updated to match and produce a success.
interpreter/Interpreter/Wasm/Examples/EarlyBr.lean — a minimal example with a function whose body is [.localGet 0, .br 0] returns its argument.
interpreter/Interpreter/Wasm.lean — imports the new example.

This resulted in 30 failing tests in testsuite due to "Unexpected break targeting function" to now pass in br_if.wast, br_table.wast, func.wast, and unwind.wast.


interpreter/Interpreter/Wasm/Semantics/Lemmas.lean
```diff
@@ -258,7 +258,7 @@ theorem run_eq
                     (f.toLocals (args.take f.numParams)) f.body with
            | .Fallthrough st s => .Success (s.values ++ args.drop f.numParams) st
            | .Return st vs     => .Success vs st
-           | .Break _ st _     => .Trap st "Unexpected break targeting function"
+           | .Break _ st s     => .Success (s.values ++ args.drop f.numParams) st
```
mfornet: This should only work if `k = 0`, if it is larger than 0 it should be considered invalid, so 
```suggestion
           | .Break 0 st s     => .Success (s.values ++ args.drop f.numParams) st
           | .Break (k + 1) st s     => .Invalid "Unexpected break targeting scope out of function"
```

mfornet: It would be great to add yet another Example that show case the Invalid path is triggered properly

mfornet (commented):
Thanks for contributing and noticing this, I'm sure the interpreter is full of similar quirks.

As I reviewed your PR I noticed a few artifacts created by Claude during the development that I'll be cleaning (for example the fact that you needed to fix your code in two places rather than one)

interpreter/Interpreter/Wasm/Semantics/Lemmas.lean
mfornet: Already updated the code, such that this change only needs to land in one place rather than two.

waterball01:
I updated it to work only if k = 0 like you said and added the extra example.

interpreter/Interpreter/Wasm/Semantics/Lemmas.lean
```diff
@@ -250,10 +250,13 @@ theorem run_eq
...
-         | .Break _ st _     => .Trap st "Unexpected break targeting function"
+         | .Break 0 st s     =>
+           .Success (s.values.take f.results.length ++ callerRemainder) st
+         | .Break (_+1) _ _  =>
+           .Invalid "Unexpected break targeting scope out of function"
          | .Invalid msg      => .Invalid msg
-         | .Trap st msg      => .Trap st msg
-         | .OutOfFuel        => .OutOfFuel) := by
+         | .OutOfFuel        => .OutOfFuel
+         | .Trap st msg      => .Trap st msg) := by
```
mfornet: Out of curiosity, why changint this two lines?

waterball01: I had an error when running lake build and I saw in Semantics.Lean it was the other order so I swapped the two in Lemmas and it fixed it.

PR #18, opened 2026-05-29
Multivalue

waterball01:
I picked this from the roadmap, and I added Examples/MultiValue.lean as a minimal end to end check, specifically swap_runs is a concrete run via native_decide, swapSpec and pairBlockSpec are FuncSpecs for a two result function and a block with resultArity = 2, callsSwapSpec is chechiking when caller consumes both results through wp_call_cons, and multiValueBlockTypeDecodes is a decoder regression for (block (type $pair)). 

In Decoder/Wat.lean, block/loop/if parsing now resolves (type N) / (type $sig) against the module’s type table, the same lookup as function headers, threaded through Ctx.resolveBlockType and skipBlockType.

Changes:
interpreter/Interpreter/Wasm/Examples/MultiValue.lean — new examples and proofs above.
interpreter/Interpreter/Wasm.lean — import MultiValue so they build with the package.
interpreter/Interpreter/Wasm/Decoder/Wat.lean — honour (type N) on structured control-flow headers.
Tested with lake build from interpreter/. I used AI to help plan the work and navigate the codebase (especially Lean proof syntax in the example file) but reviewed the diffs; the example proofs follow the same FuncSpec.of_wp_body / wp_run pattern as the existing examples.

### DIFF
--- a/interpreter/Interpreter/Wasm.lean
+++ b/interpreter/Interpreter/Wasm.lean
@@ -32,6 +32,7 @@ import Interpreter.Wasm.Examples.MemGrow
 import Interpreter.Wasm.Examples.MemFill
 import Interpreter.Wasm.Examples.MemCopy
 import Interpreter.Wasm.Examples.GlobalCounter
+import Interpreter.Wasm.Examples.MultiValue
 
 /-! # Wasm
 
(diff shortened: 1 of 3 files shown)


waterball01:
I realized I messed up my rebasing, so it included commits already merged from #5, I will quickly fix this and open a new one

PR #35, opened 2026-06-02
Fix Memchr proof: swap arg order so entry reversal gives correct locals

Isabelle9999:
The Memchr proof was broken (excluded from build), because TerminatesWith passed args as [ptr, len, needle], but entry reverses them, so the runtime had [needle, len, ptr], mismatching the invariant and making the initial step unprovable (needle = ptr).


mfornet:
Is it possible that you were using an old version of the project. It works for me locally, and this PR fails, I'll close it, but feel free to reopen it if you still see this problem.

PR #36, opened 2026-06-02
Replace trivial spec with formal TerminatesWith statement

Isabelle9999:
Replaced the spec (placeholder Prop := True) with a TerminatesWith statement that captures the informal spec from the doc comment: the function terminates with an empty result stack, and afterwards every u32 word at ptr + 4*i equals the original word at ptr + 4*(len-1-i). Args passed in reversed order ([len, ptr]) to match the entry convention confirmed in XorSum. Memory bounds side-condition follows the same hmem pattern as XorSum and Memchr. 

Branched off fix/multivalue-store-api.


mfornet:
Thanks for the contribution and welcome to the project. We are thinking about how to update the spec written in terms of the memory, since the current approach is not very readable or maintable long term. Even for this simple instances gets a bit complicated.

PR #46, opened 2026-06-06
reverse_inplace: prove in-place array reversal correct

mfornet:
## What

Discharges the previously deferred `reverse_inplace` specification. This is the first corpus example that **mutates** linear memory and whose postcondition relates the *final* memory to the *initial* store, for every `0 ≤ i < len`:

```
st'.mem.read32 (ptr + 4*i) = initial.mem.read32 (ptr + 4*(len-1-i))
```

## Changes

- **New reusable frame lemmas** — `codelib/CodeLib/Mem.lean`:
  - `Mem.read32_write32_self` — read-your-writes (the little-endian split/reassemble round-trip, settled by `bv_decide`).
  - `Mem.read32_write32_of_disjoint` — frame rule: a read whose 4-byte window misses the write's window is unaffected.
  - `Mem.write32_pages` / `Mem.write32_bytes_outside` — supporting facts.
  - Wired into the `CodeLib` umbrella import.
- **The loop proof** — `programs/lean/Project/ReverseInplace/Spec.lean`. The invariant after `n` swaps: memory is reversed on the end-segments `[0,n) ∪ [len-n, len)` and untouched in the middle. The continuation test `c <ᵤ d-1` is shown to be exactly `n+1 < ⌊len/2⌋`, so the loop exits right after the last swap; at exit the middle is empty (even `len`) or the fixed center `i = len-1-i` (odd `len`), closing the postcondition.

## Spec change a reviewer should note

The side condition is strengthened from the per-element-mod form used by `xor_sum`/`memchr` to a contiguous, non-wrapping bound:

```lean
(_hbound : ptr.toNat + 4 * len.toNat ≤ initial.mem.pages * 65536)   -- buffer fits in memory
(_hwrap  : ptr.toNat + 4 * len.toNat ≤ 4294967296)                  -- buffer doesn't wrap 2^32
```

This is **necessary, not cosmetic**: the old per-element form permits two distinct word indices to alias the same physical address under 32-bit wraparound (e.g. `ptr=0, len=2³⁰+1`), which would make in-place reversal incorrect. Reads-only specs didn't care; reverse must know the words are pairwise non-overlapping. Happy to reshape the precondition if you'd prefer a different formulation.

## Verification

- `lake build` passes for the whole `programs` package (no `sorry`, no warnings).
- Axiom check on `reverse_inplace_correct` is clean: `propext`, `Classical.choice`, `Quot.sound`, plus `bv_decide` reflection certificates (same trust class as the `native_decide` this project already uses).


mfornet:
Closing: main (#43) converted `reverse_inplace` to the equivalence-check (`check`) shape after this branch's base, which obsoletes the direct-spec proof this PR contained. Abandoning per decision to take main wholesale. The branch has been reset to match main.

PR #62, opened 2026-06-11
add float wp lemmas 

Isabelle9999:
Adds `@[simp, wp_simp]` WP rewrite lemmas for every float instruction that has an `execOne` case but no `wp_*_cons` theorem.

PRs #56/#57 added float execution semantics and decoder wiring. This completes the proof infrastructure so `wp_run` can step through float instructions in proofs.

76 lemmas covering:
- constants (f32Const, f64Const)
- arithmetic (add/sub/mul/div/min/max/copysign, both widths)
- unary (abs/neg/sqrt/ceil/floor/trunc/nearest, both widths)
- comparison (eq/ne/lt/gt/le/ge, both widths, results as i32)
- memory (f32Load/f64Load/f32Store/f64Store)
- int-to-float conversions (f32ConvertI32S/U, f64ConvertI64S/U, etc.)
- float-to-int trapping truncations (i32TruncF32S/U, etc.)
- float-to-int saturating truncations (i32TruncSatF32S/U, etc.)
- demote/promote (f32DemoteF64, f64PromoteF32)
- reinterpret (all four directions)

Every proof is `by wp_atomic`. One file changed, `lake build` passes, zero sorries.

Generated with Claude Code

### DIFF
--- a/interpreter/Interpreter/Wasm/Wp/Atomic.lean
+++ b/interpreter/Interpreter/Wasm/Wp/Atomic.lean
@@ -958,4 +958,600 @@ macro "wp_atomic" : tactic => `(tactic|
        wp m rest Q { st with dataSegments := st.dataSegments.set i none } s env) := by
   wp_atomic
 
+/-! ## float constants -/
+
+@[simp, wp_simp] theorem wp_f32Const_cons :
+    wp m (.f32Const v :: rest) Q st s env ↔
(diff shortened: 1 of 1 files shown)


mfornet:
thanks for the contribution, this looks great. Going forward it would be great to add some programs that use these theorems, this one are good rewrite lemmas.

PR #63, opened 2026-06-11
add table write instructions (table.set/grow/fill/copy/init, elem.drop)

Isabelle9999:
Add six table instructions that were missing from the interpreter:
table.set, table.grow, table.fill, table.copy, table.init, and elem.drop. 

Each instruction is wired through the full stack: constructor in the Instruction type, execution semantics, WP rewrite lemma for proofs, decoder support for WAT files, and verifier emitter arm.

Includes a TableSet.lean example with round-trip write/read proofs.

Why:  these were the last table operations the decoder recognized but
could not execute. Programs using table mutation (grow, fill, copy, init from element segments) would fail at runtime.

lake build passes at 2993 jobs, zero sorries.

Generated with Claude Code

### DIFF
--- a/interpreter/Interpreter/Wasm/Decoder/Wat.lean
+++ b/interpreter/Interpreter/Wasm/Decoder/Wat.lean
@@ -1307,6 +1307,42 @@ private def stripQuotes (s : String) : String :=
     ((s.drop 1).dropEnd 1).toString
   else s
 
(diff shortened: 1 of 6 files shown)


Isabelle9999:
Closing — this accidentally included commits from other PRs, will reopen with a clean branch

PR #72, opened 2026-06-15
codelib: add memory growth and data segment initialization lemmas

Isabelle9999:
Add reusable memory lemmas to codelib

grow_preserves / grow_read8: after memory.grow, all bytes are unchanged (grow only updates the page count). Useful for proofs
involving dynamic allocation (Vec, String, any heap use).

data_segment_initialized / data_segment_initialized_read8: after writeBytesFrom, reading from the target range returns the segment bytes. Useful for proving initial memory state after module instantiation.

New file (CodeLib/Memory.lean), no existing code changed. lake build passes.

Generated with Claude Code

### DIFF
--- a/codelib/CodeLib/Memory.lean
+++ b/codelib/CodeLib/Memory.lean
@@ -1,0 +1,41 @@
+import Interpreter.Wasm
+
+namespace Wasm
+
+-- grow only updates `pages`; the byte function is unchanged at every address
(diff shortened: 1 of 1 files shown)


mfornet:
Let's avoid PRs without "evidence" of the usefulness of the inotrduced API. I recommend we base these APIs on real examples and programs that we create and use that API. I'll close it for now, but feel free to reopen if you add new examples.

PR #81, opened 2026-06-18
near/multisig: feasibility checkpoint — real multisig wasm elaborates as a Lean module

mfornet:
## What this is

A **feasibility checkpoint** for verifying NEAR `core-contracts`, starting with **multisig**. It commits one auto-generated file — `programs/lean/Project/Near/Multisig/Program.lean` — which is the *real* core-contracts `multisig` contract (near-sdk 2.0.0 / edition 2018) compiled to wasm and transpiled to a Lean `Wasm.Module`, proving the module **elaborates end to end**.

This is **not** a finished verification target. There are no specs or proofs yet.

## Why

Before investing in a `near/` verifier restructure and a NEAR-SDK reasoning layer, we needed to answer whether the real (not reimplemented) near-sdk multisig can even be (a) built, (b) decoded, and (c) elaborated by the Lean interpreter at its true scale. This PR is the evidence that it can.

## Feasibility findings (spike)

- **Build:** compiles on a modern rustc from the pinned `Cargo.lock` (no old-toolchain wrangling); 300 KB wasm.
- **Host ABI:** all 48 `env` imports the contract declares are already covered by `codelib/CodeLib/Near`.
- **Decode:** the verifier's wat decoder accepts the 111k-line / 3.4 MB wat (~2.6 s) — but needs `ulimit -s 65520`; the parser isn't tail-recursive and overflows the default 8 MB stack on modules this size.
- **Elaboration:** `lake build Project.Near.Multisig.Program` → exit 0, but **~9.7 min wall / ~7.7 GB RSS** for the single module.

## Module shape

903 defined functions, 48 imports, 11 exports (the contract methods), 58 types, 1 table + 1 element segment (`call_indirect` is used), 3 globals, 1 memory. Emitted file: ~120k lines / 4.1 MB.

## What reviewers should know

- **CI does not build this file.** It is intentionally *not* imported by `Project.lean`, so `lake build` (main CI) and `verifier check` (freshness CI) both skip it — the same mechanism the repo already uses to exclude the known-broken `Project/Memchr/Spec.lean`. The 7.7 GB cost is incurred only by explicitly running `lake build Project.Near.Multisig.Program`.
- It carries `set_option maxHeartbeats 4000000` (on top of the emitter's `maxRecDepth`) so it elaborates.
- It's a hand-placed monolithic probe produced outside the normal `verifier emit` pipeline; the path predates the planned `near/` restructure.

## Follow-ups

- A **split-module emitter** (chunk func bodies across files; assemble the module separately) so elaboration is incremental and CI-friendly.
- The **`near/` verifier restructure**: `core-contracts` as a `vendor/` submodule, nested crate discovery, externally-built wasm, `Project.Near.<Pascal>` modules.
- The **NEAR-SDK reasoning layer** (Borsh state (de)serialization, number→JSON, overflow-branch discharge, `UnorderedMap` layout) and the first general spec for a no-arg view method (`get_request_nonce` / `get_num_confirmations`).


mfornet:
Closing this, as is we can't really merge it or work with it, I'm working on a new way to deal with "large" wasm modules

PR #82, opened 2026-06-18
RustStd pipeline: reusable u64 count_ones / abs_diff proofs + cross-crate reuse

El3ssar:
## What

First slice of a **Rust std verification pipeline**: compile a std method to wasm at `opt-level 0`, transpile to Lean, and prove its behaviour once as a *reusable* theorem in `CodeLib`, then reuse that proof from other crates.

Two exemplars, one of each kind of std API:
- **`count_ones`** — Group A (compiler intrinsic; inlined at use sites).
- **`u64::abs_diff`** — Group B (a real, separately-compiled leaf function invoked via `call`).

## How it's structured

- **`CodeLib/RustStd/Frame.lean`** — opt-0 calling-convention lemmas proved once and shared: store/load round-trips (`read{32,64}_write{32,64}_same`, via `bv_decide`), stores preserve page count, `popcnt` fits in 32 bits, frame-pointer no-underflow. (At `-O0` every function spills its result through linear memory, so every proof needs these.)
- **`CodeLib/RustStd/U64/{CountOnes,AbsDiff}.lean`** — **body-parametric, store-parametric** `TerminatesWith` theorems: they hold for the exact body at *any* index in *any* module given a stack-pointer global and enough memory. `AbsDiff` additionally carries a trailing stack frame (`rest`) and **frame conditions** (globals + page count unchanged) so it composes under nested calls.
- **Per-crate `Spec.lean`** — three-line instantiations; the function index, globals, and memory bounds are discharged by `rfl` / `decide`.

## The reuse demonstration

`rust_consumer` is a structurally different program — 2-D **Manhattan distance** (`manhattan px py cx cy = abs_diff px cx + abs_diff py cy`). It shares no shape with the demo crate, yet its `abs_diff` body is emitted byte-for-byte identically (only the mangled name's hash differs). `manhattan_spec` is proved entirely by reusing `absDiff_terminates` **twice** through the interpreter's `wp_call_tw` rule — the second call relying on the frame conditions to re-establish its no-trap hypotheses. Zero re-proving of the function body.

## Notes

- All new modules build clean against `main`; no `sorry`s, no warnings.
- Corpus profile is `opt-level = 0`, `lto = false` (byte-stable std output) — matches the recent `num_integer` migration on `main`.
- Draft: this is the proof-of-concept slice; the next steps are a written playbook + a coverage backlog to scale this across the std library.


programs/lean/Project/RustStd/Spec.lean
```diff
@@ -0,0 +1,34 @@
...
+`CodeLib.RustStd.U64`. The hypotheses (function index, stack-pointer
+global, memory bounds) are all discharged by `rfl` / `decide`.
+-/
+
+namespace Project.RustStd.Spec
+
+open Wasm Wasm.RustStd.U64
+
+/-- `count_ones` (export index 1) — Group A. Reuses `countOnes_terminates`. -/
+@[spec_of "rust-exported" "rust_std::count_ones"]
```
mfornet, line 18: **`@[spec_of]` is on a `theorem`, not a `def : Prop` + `@[proves]` (extractor contract).** `EXTRACT.md` §P4 says a `FormalSpec` is a `def Name : Prop` carrying `@[spec_of]`, and §P5 says a `Verification` is a `theorem` carrying `@[proves SpecName]`. The `is_even` template and `num_integer` both follow this split (`def IsEvenSpec : Prop` + `@[proves] theorem is_even_correct`). As written, the extractor would likely emit **zero FormalSpecs and zero Verifications** for these crates — the proofs are correct and build, but the verifier artifact wouldn't link them. Same issue on `abs_diff_spec` (L27) and `manhattan_spec` in RustConsumer.

Suggest splitting each into `@[spec_of …] def …Spec : Prop` + `@[proves …] theorem`, mirroring `num_integer`. This is the one I'd treat as blocking.

programs/lean/Project/RustStd/Spec.lean
```diff
@@ -0,0 +1,34 @@
...
+theorem count_ones_spec (env : HostEnv Unit) (v : UInt64) :
+    TerminatesWith env «module» 1 «module».initialStore [.i64 v]
+      (fun _ rs => rs = [.i32 (UInt32.ofNat (popcnt64 64 v 0))]) :=
+  countOnes_terminates env «module» 1 «module».initialStore v 1048576
+    rfl rfl (by decide) (by decide) rfl
+
+/-- The inner `core::num::<impl u64>::abs_diff` (function index 0) — Group B.
+Reuses `absDiff_terminates`. This is the body that other crates will also
+contain and call. -/
+@[spec_of "rust-exported" "rust_std::abs_diff"]
```
mfornet, line 29: **`rust_std::abs_diff` names a non-existent export.** There is no `abs_diff` symbol in this crate — the export is `u64_abs_diff` — and this proof targets `func0`, the std-internal `core::num::<impl u64>::abs_diff` (index 0), which isn't exported at all. Per `EXTRACT.md` §P4 a same-crate `rust-exported` target that doesn't resolve emits `unresolved_spec_of_target`. Either point the spec at the real exported wrapper (`u64_abs_diff` / `func3`) or, since the docstring already calls this "the inner abs_diff," tag it `rust-internal` (e.g. `core::num::abs_diff`).

programs/lean/Project/RustStd/Spec.lean
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Project.RustStd.Spec
+
+open Wasm Wasm.RustStd.U64
+
+/-- `count_ones` (export index 1) — Group A. Reuses `countOnes_terminates`. -/
+@[spec_of "rust-exported" "rust_std::count_ones"]
+theorem count_ones_spec (env : HostEnv Unit) (v : UInt64) :
+    TerminatesWith env «module» 1 «module».initialStore [.i64 v]
+      (fun _ rs => rs = [.i32 (UInt32.ofNat (popcnt64 64 v 0))]) :=
```
mfornet, line 21: _Minor:_ this spec restates the interpreter's own `popcnt64` helper (from `Semantics.lean`), so it's somewhat circular. Compare `num_integer`, whose spec uses the independent `Nat.gcd a.toNat b.toNat`. The `abs_diff` spec below is genuinely independent (`if a < b then b - a else a - b`), which is the right standard — consider relating `count_ones` to an independent popcount notion too. Consistent with project philosophy either way.

programs/rust/rust_std/src/exports.rs
```diff
@@ -0,0 +1,22 @@
+//! Wasm ABI surface for `rust_std` — the two reuse-experiment exemplars.
+//!
+//! * `count_ones` — Group A (compiler intrinsic; inlined at every use site).
+//! * `u64_abs_diff` — Group B (a real, separately-compiled leaf function,
+//!   `core::num::<impl u64>::abs_diff`, invoked here via `call`). This is
+//!   the body we prove once and expect to reuse across crates.
+
+#[unsafe(no_mangle)]
+fn count_ones(v: u64) -> u32 {
```
mfornet, line 9: **Missing `pub extern "C"`.** `EXTRACT.md` §P2 and the `ExportedFunction` source rule key on `#[unsafe(no_mangle)] pub extern "C" fn` in `exports.rs`. The wasm does export `count_ones` (and `u64_abs_diff` at L14 — same issue), but the extractor's source scan won't recognize them as exports, so even a correctly-tagged `rust-exported` spec would fail to resolve. The template, `is_even`, `num_integer::gcd_u64`, and `entrypoint`/`manhattan` in this very PR all use the full `pub extern "C"` form — make these match.

programs/rust/rust_std/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+mod exports;
+
+/// Starting-point implementation for `rust_std`. Replace the body with the
+/// real logic you want to verify; keep the pure function here and expose it
+/// across the wasm ABI in `exports.rs`.
+pub fn rust_std(n: i32) -> bool {
```
mfornet, line 6: _Cleanup:_ this is the verbatim unmodified `CRATE_NAME` scaffolding from `verifier/template/crate/src/lib.rs` (comment and `n % 2 == 0` body included). Nothing references it and it doesn't appear in the generated module. Remove it — `rust_consumer/src/lib.rs` (just `mod exports;`) is the right shape here, since the "pure functions" in this crate are the std methods themselves.

codelib/CodeLib/RustStd/Frame.lean
```diff
@@ -0,0 +1,91 @@
...
+  have e3 : a.toNat + 3 ≠ a.toNat := by omega
+  have e21 : a.toNat + 2 ≠ a.toNat + 1 := by omega
+  have e31 : a.toNat + 3 ≠ a.toNat + 1 := by omega
+  have e32 : a.toNat + 3 ≠ a.toNat + 2 := by omega
+  simp only [e1, e2, e3, e21, e31, e32, ↓reduceIte]
+  bv_decide
+
+/-- Reading a 64-bit word back from the address it was just written to
+returns the stored value. -/
+@[simp] theorem Mem.read64_write64_same (m : Mem) (a : UInt32) (v : UInt64) :
```
mfornet, line 45: _Minor:_ this duplicates the local `read64_write64_same` already in `num_integer/Spec.lean` (same `bv_decide` proof). CodeLib is the right home — nice — so consider a follow-up that slims `num_integer` to reuse it. Also note these round-trip / `pages` lemmas are added as global `@[simp]` and CodeLib is widely imported; they're well-behaved rewrites, but worth a sentence in the PR description since it changes the global simp set.

programs/lean/Project/RustConsumer/Spec.lean
```diff
@@ -0,0 +1,53 @@
...
+own no-trap hypotheses after the first call mutated memory.
+-/
+
+namespace Project.RustConsumer.Spec
+
+open Wasm Wasm.RustStd.U64
+
+set_option maxRecDepth 4096 in
+/-- `manhattan px py cx cy = abs_diff px cx + abs_diff py cy`. -/
+@[spec_of "rust-exported" "rust_consumer::manhattan"]
```
mfornet, line 36: Same `@[spec_of]`-on-`theorem` issue as the RustStd specs — should be `def …Spec : Prop` + a separate `@[proves]` theorem per `EXTRACT.md` §P4/§P5. (The export form here is correct: `manhattan` is `#[unsafe(no_mangle)] pub extern "C"` and the target resolves.)

mfornet (commented):
Review of the RustStd pipeline slice. I built both affected packages locally — `codelib` (`Frame`, `U64.CountOnes`, `U64.AbsDiff`) and `programs/lean` (`RustStd.Spec`, `RustConsumer.Spec`) — and both compile clean with no `sorry` and no warnings. The proof content and the central cross-crate reuse claim are sound, and the body-parametric-theorem + frame-conditions design is exactly right.

The inline comments below are all in the **spec/export metadata layer**, not the proofs. Several contradict the repo's own committed contract (`verifier/EXTRACT.md`) and the only precedent (`num_integer`) / the `is_even` template, so as written the extraction pipeline would likely not pick these specs up. Acknowledging this is a draft — flagging what to fix before it lands. None require touching the proofs.

mfornet:
Closing in favor of #84 and #85

PR #85, opened 2026-06-18
Reusable u64 std corpus (abs_diff) + cross-crate reuse demo

El3ssar:
Proves a Rust std function once in CodeLib and reuses the proof in an unrelated program.

**Foundation (proved once):**
- `CodeLib/RustStd/Frame.lean` — opt-level-0 calling-convention lemmas (store/load round-trips, stores preserve page count, `popcnt` width bound, frame-pointer no-underflow).
- `CodeLib/Entry.lean` — `Returns` (a readable wp post-condition) and `TerminatesWith.of_returns_wp` (bridges a wp body lemma to the `TerminatesWith` the call rule needs).

**The function:**
- `CodeLib/RustStd/U64/AbsDiff.lean` — `absDiff_wp`, stated in **wp form about the body** (no `TerminatesWith`, no module/host hypotheses), for the inner `core::num::<impl u64>::abs_diff`.
- `rust_u64` crate (one crate per primitive type) exports `abs_diff`; its per-crate spec is the two-declaration form (`@[spec_of "rust-internal" "core::num::abs_diff"] def` + `@[proves] theorem`) that bridges `absDiff_wp`.

**The reuse demo:**
- `total_variation` crate computes `|a-b| + |b-c|` — structurally unlike `rust_u64` — and its proof reuses `absDiff_wp` twice via `wp_call_tw`, with **no CodeLib change and the body never re-proven**.

Note: `Frame` introduces a few global `@[simp]` round-trip/`pages` lemmas; `num_integer` has a local `read64_write64_same` that can be slimmed later to reuse Frame's (follow-up).


mfornet (commented):
**Verdict: sound and well-executed.** The "prove-once-in-CodeLib, reuse-across-crates" pattern this demonstrates is genuinely valuable and the mechanism is clean. CI is green (`Build & verify proofs`), there are no `sorry`/`admit`/`axiom` escapes, and the specs faithfully capture the Rust semantics. Comments below are about design/maintainability, not correctness.

## What's good

- **The core contribution lands.** `absDiff_wp` is stated in wp/body form with no module or `TerminatesWith` baggage, `of_returns_wp` bridges it to the call rule, and `TotalVariation/Spec.lean` reuses it twice via `wp_call_tw` with the body never re-proven. That's exactly the claim, and the demo crate is structurally unlike the source crate, so it's a real test of reuse.
- **Specs match Rust.** `abs_diff` → `if a < b then b - a else a - b` (unsigned `<`, matches `u64`); `total_variation` → `(…) + (…)` with `+` wrapping, matching `wrapping_add`.
- `Returns` is a clean, readable post-condition abstraction, and `of_returns_wp`'s bound-shifting (`args.drop f.numParams` re-appended at the bridge, not in the body lemma) is the right place for it.
- Convention respected: `lib.rs` is just `mod exports;`, all `unsafe` lives in `exports.rs`.

## Substantive comments

**1. Duplication is wider than the PR note admits.** The description flags only `num_integer`'s local `read64_write64_same` as a future dedup. But `Frame.lean` actually overlaps **four** lemmas with `programs/lean/Project/NumInteger/Spec.lean`: `read64_write64_same` (line 51), `read32_write32_same` (line 114), and — already `@[simp]` in *both* places — `write64_pages` (line 160) and `write32_pages` (line 163). So after this PR there are two identical global `@[simp]` rules for `(m.write{32,64} a v).pages`. Harmless (same LHS/RHS, terminating, CI green), but the follow-up should slim all four, not just one, and `num_integer` should import Frame's.

**2. `@[simp]` lemmas go global.** Frame is imported by `CodeLib.lean`, so its four `@[simp]` lemmas now live in the default simp set for the entire downstream corpus. The read-after-write-same and pages lemmas are good, confluent simp candidates and the full corpus rebuilt green, so risk is low — just flagging that this is a repo-wide decision riding in on a feature PR, worth a conscious ack.

**3. `absDiff_wp`'s post-condition preserves only `globals` and `mem.pages`, not memory contents.** It says nothing about the caller's memory outside the spilled frame `[sp-16, sp)`. That's sufficient for `total_variation` (nothing reads the scratch region between calls), but a future caller that needs "this call didn't clobber my data at address X" can't get it from this lemma as stated. Consider whether the reusable body theorem should also assert frame-locality (memory unchanged outside the frame) so it composes in more situations.

**4. Half of Frame is unexercised by this PR.** `absDiff_wp` only uses `Mem.read64_write64_same`, `Mem.write64_pages`, and `UInt32.toNat_sub_of_le`. The two `popcnt64_*` lemmas and the two 32-bit lemmas aren't touched by any proof here. Defensible for a "foundation" module, but nothing yet validates that `popcnt64_lt_2pow32` is the right shape for the `i32.wrap_i64 ∘ i64.popcnt` obligation it's meant for — it's speculative until a real popcnt proof consumes it.

## Minor nits

- `rust_u64/src/exports.rs` and `total_variation/src/exports.rs` are missing a trailing newline.
- `total_variation` has no `entrypoint` wrapper, contradicting the rust_u64 convention the guide states ("a single `entrypoint` that calls all of them so nothing is dropped"). It's fine here because `total_variation` is itself a root export and won't be DCE'd — but `docs/std_guide.md` should say *why* the entrypoint is unnecessary in the reuse case, since it otherwise looks like an inconsistency.
- `docs/std_guide.md` has stray double blank lines and the heading `## Example prove u64::abs_diff` reads awkwardly (missing punctuation, e.g. `## Example: prove u64::abs_diff`).
- The hard-coded inner index `0` and stack pointer `1048576` are brittle across re-emits; the guide already calls out the index as the one number to bump, which is the right mitigation.

None of these block merge. The dedup (#1) and the memory-contents post-condition (#3) are the two worth a sentence in the PR or a tracked follow-up before the corpus grows much further.

mfornet (approved):

PR #86, opened 2026-06-19
codelib: drop duplicate UInt{32,64}.toNat_sub_of_le lemmas

El3ssar:
## What

Removes two hand-written lemmas that duplicate the Lean standard library:

- `Wasm.UInt32.toNat_sub_of_le` in `codelib/CodeLib/RustStd/Frame.lean`
- `Wasm.UInt64.toNat_sub_of_le` in `codelib/CodeLib/UInt64.lean`

Both were verbatim re-proofs of lemmas Lean core already ships in `Init.Data.UInt.Lemmas` (always in scope, no Mathlib needed), with **identical** signatures:

```
UInt32.toNat_sub_of_le : ∀ (a b : UInt32), b ≤ a → (a - b).toNat = a.toNat - b.toNat
UInt64.toNat_sub_of_le : ∀ (a b : UInt64), b ≤ a → (a - b).toNat = a.toNat - b.toNat
```

## Why it's safe

Because the core lemmas share the same root name, every call site rebinds to them automatically once the local copies are removed — **no call-site edits required**. The affected references (`AbsDiff.lean`, `UInt64.lean`'s internal Stein lemmas, and `NumInteger/Spec.lean`) all resolve to core.

Verified by building both `codelib` (3001 jobs) and `programs/lean` (3009 jobs) with the duplicates gone — green.

## What's kept

The four `Mem.read/write` framing lemmas in `Frame.lean` stay: `Mem` is the repo's own type, those lemmas have no stdlib equivalent, and removing them breaks the consumers (confirmed). The `Frame.lean` docstring is updated to note that the frame-pointer no-underflow fact now comes from core.

### DIFF
--- a/codelib/CodeLib/RustStd/Frame.lean
+++ b/codelib/CodeLib/RustStd/Frame.lean
@@ -16,8 +16,10 @@ form that does not care about the concrete spill address:
 * `Mem.read{32,64}_write{32,64}_same` — read-after-write at the same
   address returns the value (no bound on the address needed).
 * `Mem.write{32,64}_pages` — a store leaves the page count unchanged.
-* `UInt32.toNat_sub_of_le` — frame-pointer arithmetic (`sp - 16`) without
-  underflow, needed for the in-bounds (no-trap) obligation.
+
+The in-bounds (no-trap) obligation also needs `sp - 16` not to underflow;
+that is `UInt32.toNat_sub_of_le`, which Lean core already provides
+(`Init.Data.UInt.Lemmas`), so it is used directly rather than re-proved here.
 
 The four `Mem.*` lemmas are intentionally global `@[simp]` — confluent,
 terminating rewrites used corpus-wide. Op-specific lemmas (e.g. `popcnt`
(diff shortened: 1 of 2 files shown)


El3ssar (commented):
**Reviewed — correct, well-scoped, safe to merge.** I re-derived each claim independently rather than taking the description at face value:

- **Core lemmas exist and match.** `Init/Data/UInt/Lemmas.lean:148` generates, per integer type, `@[simp] protected theorem UIntN.toNat_sub_of_le (a b) : b ≤ a → (a - b).toNat = a.toNat - b.toNat`. Same statement as the deleted local copies; the only difference is core writes the hypothesis as an arrow vs. the local `(h : b ≤ a)` binder — identical Pi type, so every application (`… sp 16 hle`, partial `… x y`) still typechecks.
- **Name resolution rebinds cleanly.** The deleted decls lived in `namespace Wasm`, i.e. `Wasm.UInt{32,64}.toNat_sub_of_le`, shadowing core for explicit references. All five call sites across the three files (`RustStd/U64/AbsDiff.lean`, `UInt64.lean`'s Stein lemmas, `NumInteger/Spec.lean`) use the fully-qualified name under `namespace Wasm` / `open Wasm`, so they fall back to core once the locals are gone. `protected` on the core version doesn't bite here precisely because the references are fully qualified.
- **No simp-set perturbation.** Core's `toNat_sub_of_le` is `@[simp]` and was always in scope; the deleted duplicates were differently-named, non-`@[simp]` decls. So removing them leaves the simp set untouched — no risk of a `simp`-driven proof elsewhere shifting.
- **Build is green.** Applied both files from this branch onto a clean tree and rebuilt: `codelib` → 3001 jobs, `programs/lean` → 3009 jobs, both successful. Matches the description.

Docstring edit in `Frame.lean` is accurate. No changes requested.

(Self-authored PR, so this is a `COMMENT` review rather than an approval; CODEOWNERS still requires @mfornet to approve before merge.)

mfornet (approved):

PR #87, opened 2026-06-19
u64: prove `add` in CodeLib + `rust_u64_tests` reuse crate

El3ssar:
Adds `u64::add` to the CodeLib corpus and its reuse test, via the std-verification pipeline (`docs/std_guide.md`). Follows PR #85 (`abs_diff` + `total_variation`).

## What's proven
- **`CodeLib/RustStd/U64/Add.lean`** — `add_wp`, the reusable wp-form body theorem: `u64::add a b = a + b` (the wrapping add; `+` on `UInt64` is wrapping, matching the unchecked `i64.add`). No `TerminatesWith`, no module/host hypotheses.
- **`Project/RustU64/Spec.lean`** — `AddSpec` + `add_correct`.

## Note: `add` inlines (vs `abs_diff`)
At `opt-level = 0` the `+` operator compiles to a single `i64.add` with **no inner call and no stack frame** — unlike `abs_diff`, which forwards to a real inner std function. Consequences:
- The lifted function is the **export itself** (index 2), so the spec is tagged **`rust-exported "rust_u64::add"`** (not `rust-internal`).
- The proof **drops the spill-frame `sp/hsp/hlo/hhi` hypotheses** (the function touches neither the shadow stack nor memory), keeping only the `globals`/`mem.pages` post-condition so it still composes under the `call` rule.

## Reuse tests: one accumulating crate
Instead of a crate per function, reuse demos now live in a single accumulating crate **`rust_u64_tests`** (`Project.RustU64Tests`), one test fn per std fn — mirroring how `rust_u64` accumulates the corpus, and avoiding workspace sprawl.

- `add_sum3 a b c = (a + b) + c` reuses `add_wp` **twice** through the `call` rule (`of_returns_wp` + `wp_call_tw`), re-proving nothing about `add`'s body.
- Because `+` inlines, the test uses a local `#[no_mangle] extern "C"` `add` shim whose opt-0 body is identical to `rust_u64::add` — so its emitted `Function` record is definitionally CodeLib's `addFunc`, and the `.call` reuses the theorem. (For std fns that compile to a real call, like `abs_diff`, you'd call directly.)

## Verification
All four crates prove green:
```
==> 4 crate(s), lake build OK, 0 sorry warning(s)
```
(`num_integer`, `rust_u64`, `total_variation`, `rust_u64_tests`)


El3ssar:
### Code review

No correctness or CLAUDE.md issues. Verified the substantive parts: the func-index renumbering (`add` = func2, `entrypoint` = func3, with `func3`'s body calling `.call 1`/`.call 2` for `abs_diff`/`add` respectively), the `rust-exported` spec tag, fuel-free `TerminatesWith` specs, and the `add_wp`-via-`call` reuse in `rust_u64_tests` (the shim's opt-0 body is definitionally `addFunc`, so `add_sum3` composes without re-proving). Addition's commutativity makes the `[b, a]` operand order harmless.

One non-blocking follow-up (not a blocker for this PR):

`add_wp`'s post-condition preserves only `globals`/`mem.pages`, not memory contents — even though `add` writes no memory at all and could assert full inertness. This is the same composability point raised for `absDiff_wp` in [#85's review](https://github.com/cajal-technologies/talos/pull/85#pullrequestreview-4531610519) ("the memory-contents post-condition ... worth a tracked follow-up before the corpus grows much further"). The corpus is now growing, so it may be worth deciding whether to strengthen the reusable body theorems' frame before the pattern entrenches.

https://github.com/cajal-technologies/talos/blob/31aa6fd6bd110294b4e7b016af47a3fec32f1fc6/codelib/CodeLib/RustStd/U64/Add.lean#L42-L46

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

El3ssar:
This PR was closed when its branch `u64-add` was renamed to `u64/operators`. The work continues (consolidated with sub/mul/div) in #92.

PR #88, opened 2026-06-19
u64: prove `mul` in CodeLib + `rust_u64_tests` reuse

El3ssar:
Adds `u64::mul` to the reusable CodeLib std corpus, following the same
9-step pipeline as `abs_diff` (#85) and `add` (#87).

## What

- **`CodeLib/RustStd/U64/Mul.lean`** — `rust_u64::mul(a, b) = a * b` wraps at
  `opt-level=0` to a single `i64.mul` (no inner `call`, no stack frame, no
  memory spill), so the export itself is lifted. `mulBody`/`mulFunc` copied
  verbatim; `mul_wp` states the result in independent math as `.i64 (a * b)`
  — `*` on `UInt64` is exactly the wrapping multiply `i64.mul` computes, so
  no overflow precondition is needed. Frame-less proof (`unfold; wp_run;
  simp`), mirroring `Add.lean`. The post keeps the `globals`/`mem.pages`
  frame + trailing `vs` so it composes under the `call` rule.
- **`Project.RustU64.Spec.MulSpec`** — per-crate `TerminatesWith` spec
  (export idx 3) tagged `rust-exported "rust_u64::mul"`, bridged from
  `mul_wp` via `TerminatesWith.of_returns_wp`.
- **`rust_u64_tests::mul_prod3 a b c = (a * b) * c`** — reuse test exercising
  `mul_wp` twice through the `call` rule (`of_returns_wp` + `wp_call_tw`),
  mirroring `add_sum3`. `mul` inlines, so a local shim with identical codegen
  (`== mulFunc`) is exposed and called.

## Verification

Both crates build green with **0 `sorry`**:

```
just verifier-prove rust_u64         # Project.RustU64.Spec        ✔
just verifier-prove rust_u64_tests   # Project.RustU64Tests.Spec   ✔
```


El3ssar:

PR #89, opened 2026-06-19
rust_u64: prove u64::sub (wrapping) + cross-crate reuse test

El3ssar:
## What

Adds `u64::sub` to the `rust_u64` std-verification corpus, following `docs/std_guide.md` end-to-end.

- **CodeLib theorem** — [`codelib/CodeLib/RustStd/U64/Sub.lean`](codelib/CodeLib/RustStd/U64/Sub.lean): `sub_wp` proves `u64::sub a b = a - b` in **wp form about the body**, with the `globals`/`mem.pages` frame post-condition kept so it composes under `call`.
- **Per-crate spec** — [`programs/lean/Project/RustU64/Spec.lean`](programs/lean/Project/RustU64/Spec.lean): `SubSpec` (`TerminatesWith`, tagged `rust-exported "rust_u64::sub"`), discharged from `sub_wp` via `of_returns_wp` + `.mono`.
- **Reuse test** — new accumulating crate `rust_u64_tests` with `sub_chain3 a b c = (a - b) - c`, proven by reusing `sub_wp` twice through the `call` rule.

## Why it's the frame-less template

At `opt-level = 0` (release profile, overflow checks off) the `-` operator inlines to a single `i64.sub` — no inner call, no shadow-stack frame, no memory spill. So:

- The exported `sub` **is** its own body ⇒ lifted the export itself (module index 3) ⇒ `rust-exported`, not `rust-internal`.
- `-` on `UInt64` is wrapping subtraction, exactly what unchecked `i64.sub` computes ⇒ no underflow precondition.
- The spill-frame hypotheses (`sp`/`hsp`/`hlo`/`hhi`) are dropped; only the `globals`/`mem.pages` post remains. This mirrors the `add` template.

Because `sub` inlines, the reuse test exposes a local `sub` shim with the identical body (so its emitted `Function` record is definitionally `subFunc`) and calls it, so each `.call 0` in `sub_chain3` reuses `sub_wp`.

## Verification

```
just verifier-prove rust_u64        # lake build OK, 0 sorry
just verifier-prove rust_u64_tests  # lake build OK, 0 sorry
```

Both crates build green with no `sorry`.

> Note: the module index `3` in both `Spec.lean` files is correct against the current `main` (abs_diff only). If `add`/`mul` land first, the shared `rust_u64` module renumbers and the index must be bumped after re-emit.


El3ssar:

PR #90, opened 2026-06-19
u64::div: reusable CodeLib theorem (tail-generic) + cross-crate reuse test

El3ssar:
## What

Verifies `u64::div` end-to-end through the std pipeline (`docs/std_guide.md`): proves `u64::div a b = a / b` once in CodeLib (`wp` form about the body) and reuses it under `call`.

## How `div` compiles (and why it's different from add/sub/mul)

At `opt-level = 0` the `/` operator is **inlined** directly into the export — no inner `call`, no stack frame, no memory spill — so this lifts the export's own index (idx `2` off main), tagged `rust-exported "rust_u64::div"`.

Unlike `+`/`-`/`*` (which wrap silently in the release profile), **division-by-zero is always checked** — it's UB in LLVM even in release. So the body is a `block` that guards the `i64.div_u` and, when `b == 0`, breaks out to a per-crate panic stub (`const <&Location>; call <panic>; unreachable`, "attempt to divide by zero"). The spec therefore carries a `b ≠ 0` precondition.

## The reusability wrinkle — tail-generic `div_wp`

The panic stub's `const`/`call` indices are **crate-specific**: the same `div` compiled into a different module gets a different tail (e.g. `rust_u64`: `const 1048600; call 56`; `rust_u64_tests`: `const 1048608; call 54`). A byte-identical-body reuse (like `add`'s shim trick) would therefore fail.

But with `b ≠ 0` that tail is **unreachable** — the guard block returns first. So `div_wp` is stated over the shared guard `block` (`divGuard`) followed by an **arbitrary `tail`**, which lets the single CodeLib theorem reuse under `call` for *any* crate's `div`. `divBody`/`divFunc` instantiate `tail` at the verbatim `rust_u64` body for the per-crate spec. This is also robust to the panic index shifting when more functions land on `main`.

## Changes

- **`codelib/CodeLib/RustStd/U64/Div.lean`** (new): `divGuard`, `divBody`, `divFunc`, and `div_wp` — `unfold divGuard Returns; apply wp_block_cons; wp_run; simp [hb]`.
- **`Project/RustU64/Spec.lean`**: `DivSpec` / `div_correct` (`rust-exported "rust_u64::div"`, idx 2, `b ≠ 0`).
- **`rust_u64_tests`** (new accumulating reuse crate + `Project.RustU64Tests`): `div_chain a b c = (a / b) / c` reuses `div_wp` twice via `of_returns_wp` + `wp_call_tw`; the shim shares `divGuard` but has its own panic tail, inferred from `func0Def.body`.
- `rust_u64/src/exports.rs`, `CodeLib.lean`, `Project.lean`, `Cargo.{toml,lock}`: wire-up.

## Verification

`just verifier-prove rust_u64` and `just verifier-prove rust_u64_tests` both build green, **0 sorry**; full `programs` lib builds green.

## Note for rebase

If add/sub/mul merge first, the `div` export index (`2`) in `Project/RustU64/Spec.lean` and `divBody`'s panic-tail literal will renumber — re-emit, bump the idx, and refresh the tail. The tail-generic `div_wp` proof itself is unaffected.

### DIFF
--- a/codelib/CodeLib.lean
+++ b/codelib/CodeLib.lean
@@ -5,6 +5,7 @@ import CodeLib.UInt32
 import CodeLib.UInt64
 import CodeLib.RustStd.Frame
 import CodeLib.RustStd.U64.AbsDiff
+import CodeLib.RustStd.U64.Div
 import CodeLib.RustStd.Option
 import CodeLib.Near.State
 import CodeLib.Near.Env
--- a/codelib/CodeLib/RustStd/U64/Div.lean
+++ b/codelib/CodeLib/RustStd/U64/Div.lean
@@ -1,0 +1,84 @@
+import CodeLib.RustStd.Frame
+import Interpreter.Wasm.Wp.Tactic
+import Interpreter.Wasm.Wp.Block
+import CodeLib.Entry
+
+/-!
+# `u64::div` — reusable body theorem
+
+`u64::div a b = a / b`. At `opt-level = 0` the `/` operator is inlined directly
+into the export: no inner `call`, no stack frame, no memory spill. Unlike `+`,
+`-`, `*` — whose overflow checks are off under the corpus's release build
+(`[profile.release]` with `overflow-checks` defaulting off), so they compile to a
+bare wrapping `i64.{add,sub,mul}` — unsigned division is **always** guarded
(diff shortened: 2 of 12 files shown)


El3ssar:

PR #91, opened 2026-06-19
decoder: evaluate global.get / extended-const global initializers

hobostay:
## Problem

Global initializers that are not a single `*.const` were mishandled in the WAT decoder:

- `global.get` of an imported global — a **wasm 1.0 core** constant expression — was silently replaced with a zero placeholder (`Wat.lean`: `| some "global.get", _ => .ok (.i32 0)`). Any global derived from an imported global got the wrong value.
- Extended-const arithmetic initializers (`i32.add`/`i32.sub`/`i32.mul`, i64 ditto) were rejected outright with `"global init expression must be i32.const or i64.const"`.

This surfaced as 8 failing spec-testsuite assertions:

```
global.wast:210-215   (z1-z6: global.get + extended-const arithmetic)
i31.wast:148          (ref.i31 of global.get in a global initializer)
linking.wast:376      (global.get of an imported global in an initializer)
```

Example (`global.wast`):
```wat
(global (import "spectest" "global_i32") i32)        ; imported global 0 = 666
(global $z1 i32 (global.get 0))                       ; expected 666, got 0
(global $z5 i32 (i32.add (global.get 0) (i32.const 42)))  ; expected 708, rejected
(global $z3 i32 (i32.sub (i32.mul (i32.const 20) (i32.const 3)) (i32.const 18)))  ; expected 42, rejected
```

## Fix

Route these initializers through the **existing** `GlobalDecl.initExpr` mechanism so `Module.runConstGlobals` evaluates them at instantiation (once imported globals are in the store) — exactly like the GC `struct.new`/`array.new*` case that already used this path. The placeholder `init` keeps the declared value type for static type-checking (`Value.toValueType`).

Detection (`initExprNeedsEval`, mutual over `Sexpr`/`List Sexpr`) scans the initializer for a `global.get` or arithmetic op, **recursing into folded forms** and covering the **flat instruction sequence** that `wasm-tools print` emits (`i32.const 20 i32.const 2 i32.mul …`). Plain const / ref / v128 initializers contain none of these and stay on their literal-folding path unchanged.

The runner now also applies `runConstGlobals` to the fresh store before running, matching the testsuite instantiation path — otherwise constant-expression globals kept their zero placeholder when run via `lake exe runner`.

## Verification

- `lake build Interpreter` — clean (all proofs + `native_decide` examples pass), including the new `Examples/GlobalInitExpr.lean` regression proof.
- `scripts/runner-smoke.sh` — all 5 checks pass.
- `just testsuite-report` (pinned `wasm-tools 1.251.0`) — regenerated `testsuite_report.txt`: **+8 pass, 0 regressions** (all 8 failing lines above flipped to `pass`; no other line changed).

The new example module decodes `(i32.sub (i32.mul (i32.const 20) (i32.const 3)) (i32.const 18))` and proves, via `native_decide`, that `runConstGlobals` writes `42` into the slot and `getG` returns `42` — locking in the behaviour against a future regression to the placeholder-zero or first-operand value.

## Scope

One logical change (global initializer evaluation), touching the decoder, the runner's store setup, the committed testsuite report, and a new example. No semantics or proof structure changed; the `initExpr`/`runConstGlobals` infrastructure was already in place for GC.

---

AI disclosure: the code in this PR was written with Claude Code. I have read and understood every change and am accountable for it; happy to address review feedback.

### DIFF
--- a/interpreter/Interpreter/Runner.lean
+++ b/interpreter/Interpreter/Runner.lean
@@ -247,7 +247,16 @@ def runOnce (a : Args) : IO UInt32 := do
   -- so we reverse here to match the call convention. (Single-arg
   -- functions are order-insensitive, which masked this for older
   -- samples.)
-  match Wasm.run a.fuel m idx (m.initialStore (α := Unit)) vs.reverse with
+  --
+  -- Evaluate constant-expression global initializers (`global.get`,
+  -- extended-const arithmetic, GC `struct.new`/`array.new*`) into the
+  -- fresh store before running, the same way the testsuite driver does
+  -- at instantiation — otherwise those globals keep their zero
+  -- placeholder. There are no imports here, so `global.get` of an
+  -- imported global is unreachable; the import pre-flight above
+  -- rejects any module that would need one.
+  let store0 := m.runConstGlobals a.fuel (m.initialStore (α := Unit)) {}
+  match Wasm.run a.fuel m idx store0 vs.reverse with
   | .Success results _ =>
     for v in results.reverse do
       IO.println (renderValue v)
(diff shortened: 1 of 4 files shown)


mfornet (approved):
Thanks for the contribution.

PR #92, opened 2026-06-19
u64: prove `add`/`sub`/`mul`/`div` in CodeLib + `rust_u64_tests` reuse

El3ssar:
Consolidates and extends the u64 std work into a single reusable corpus. Builds on the `abs_diff` corpus (#85) and supersedes the four per-operator PRs (#87 add, #89 sub, #88 mul, #90 div), then adds the rest of the operator set on top.

## What's proven

**11 operators**, each as a reusable CodeLib chunk theorem about the inlined opcode sequence, a per-crate `@[spec_of "rust-exported"]`/`@[proves]` spec in `Project/RustU64/Spec.lean`, and inline reuse tests in `Project/RustU64Tests/Spec.lean`:

- **add / sub / mul** (`a {+,-,*} b`) and **bitand / bitor / bitxor** (`a {&,|,^} b`) — each inlines at opt-0 to a single bare `i64.{add,sub,mul,and,or,xor}` (release profile ⇒ overflow checks off ⇒ wrapping, no panic branch, no frame). Frame-less chunk theorems.
- **not** (`!a`) — inlines to `i64.const 0xFFFF_FFFF_FFFF_FFFF; i64.xor`, i.e. bitwise complement (the `^ all-ones = ~~~a` step is `bv_decide`).
- **shl / shr** (`a << b`, `a >> b`, `b : u32`) — inline to the width-specific mask-extend-shift `i32.const 63; i32.and; i64.extend_u; i64.{shl,shr_u}`, i.e. a (logical) shift by `b % 64`. The `b % 64` normalisation is baked into the `shl_seq`/`shr_seq` chunk once.
- **div / rem** (`a / b`, `a % b`) — division/remainder-by-zero is always checked (it traps), so the body is a zero-divisor guard `block` around `i64.{div_u,rem_u}` plus a **crate-specific panic tail**. The reusable unit is the **tail-generic** `div_chunk`/`rem_chunk` (bare `i64.div_u`/`i64.rem_u` on stack operands, with a `b ≠ 0` hypothesis); `divBodyWp`/`remBodyWp` are crate-specific wrappers that peel the guard and reuse the chunk. Specs carry a `b ≠ 0` precondition.

`core::num::<u64>::abs_diff` (#85) remains, specced via its inner helper.

## Architecture (CodeLib reorg)

- **`CodeLib/RustStd/UInt.lean`** — type-agnostic trunk: the `UIntWasm T` class (how a `T` rides as a wasm `Value` via `toV`), `framePost`, the `BinChunk`/`UnChunk` chunk-theorem shapes (operands on the value stack, so a chunk `rw`s directly onto an inlined occurrence — no `.call`, no shim), and `binBodyWp`/`unBodyWp`, which turn any chunk into the opt-0 export-body theorem **by reusing the chunk** (the `frag` is opaque, so `wp_run` cannot bypass it).
- **`CodeLib/RustStd/U64/Basic.lean`** — the `UIntWasm UInt64` instance (`toV = .i64`) + the `toV_u64` simp lemma.
- **`CodeLib/RustStd/U64/{Add,Sub,Mul,Div,Rem,BitAnd,BitOr,BitXor,Not,Shl,Shr}.lean`** — one file per operator: the chunk theorem (discharged by the interpreter's atomic `wp_*` lemma), the `…BodyWp` export-body theorem, and a concrete-`i64` `…_seq` restatement for rewriting at inlined occurrences.

## Reuse tests (`rust_u64_tests`)

22 functions, **two structurally-distinct per operator**, each using the operator **inline** the way real client code emits it (no `#[no_mangle]` shim, no calling the op as a function):

| op | tests |
|----|-------|
| add | `add_chain = a+b+c`, `add_then_mul = (a+b)*c` |
| sub | `sub_chain = a-b-c`, `sub_then_add = (a-b)+c` |
| mul | `mul_chain = a*b*c`, `mul_then_add = a*b+c` |
| bitand | `and_chain = a&b&c`, `and_then_or = (a&b)\|c` |
| bitor | `or_chain = a\|b\|c`, `or_then_xor = (a\|b)^c` |
| bitxor | `xor_chain = a^b^c`, `xor_then_and = (a^b)&c` |
| not | `not_twice = !!a`, `not_then_xor = (!a)^b` |
| shl | `shl_then_add = (a<<n)+b`, `shl_twice = (a<<n)<<m` |
| shr | `shr_then_sub = (a>>n)-b`, `shr_twice = (a>>n)>>m` |
| div (b≠0) | `div_then_add = a/b+c`, `div_then_mul = (a/b)*c` |
| rem (b≠0) | `rem_then_add = a%b+c`, `rem_then_mul = (a%b)*c` |

Each proof discharges the inlined opcode chunk by **reusing the CodeLib chunk theorem** — `TerminatesWith.of_wp_entry_for` over the inlined body, then `rw`/`simp only` with the op's `…_seq` (or `div_chunk`/`rem_chunk`). The op's raw atomic `wp_*` lemma is **deliberately excluded** from each `simp only` set, so the CodeLib theorem is the only path through the opcode (drop it and the proof fails). This exercises the same chunk theorem that serves the called export body — and, for div/rem, reuse across crates whose panic tails differ (`call 66`/`67` in `rust_u64` vs `call 74`/`75` in `rust_u64_tests`, in the unreachable `b = 0` branch).

The `rust_u64` export bodies bridge their CodeLib body theorem to `TerminatesWith` via `of_returns_wp`; no operator body is re-proven.

## Module layout (regenerated)

Re-emitted from the unioned Rust sources. Export → `funcIdx`:

- **`rust_u64`**: `abs_diff = 1` (inner helper at func 0), `add = 2`, `bitand = 3`, `bitor = 4`, `bitxor = 5`, `div = 6`, `entrypoint = 7` (unspecced call-everything driver, returns `()`), `sub = 8`, `mul = 9`, `rem = 10`, `not = 11`, `shl = 12`, `shr = 13`.
- **`rust_u64_tests`**: `add_chain = 0`, `add_then_mul = 1`, `and_chain = 2`, `and_then_or = 3`, `div_then_add = 4`, `div_then_mul = 5`, `mul_chain = 6`, `mul_then_add = 7`, `not_then_xor = 8`, `not_twice = 9`, `or_chain = 10`, `or_then_xor = 11`, `rem_then_add = 12`, `rem_then_mul = 13`, `shl_then_add = 14`, `shl_twice = 15`, `shr_then_sub = 16`, `shr_twice = 17`, `sub_chain = 18`, `sub_then_add = 19`, `xor_chain = 20`, `xor_then_and = 21`.

All spec entry indices match these tables (both modules have `imports := []`, so entry index = `funcIdx`).

## Verification

`rust_u64` and `rust_u64_tests` build green, **0 sorry**; full `programs` lib (incl. `num_integer`, `total_variation`) builds green.

Supersedes #87, #88, #89, #90.


El3ssar:
## Deep review — verdict: ✅ sound, mergeable (one doc fix applied)

I did a deep audit of this PR: a clean rebuild from the head commit plus a multi-dimension adversarial review of semantics, spec strength, generated-module integrity, the reuse claim, and conventions. Summary below.

### Verified green (gold standard)
Rebuilt both packages from `f7a8566` against the shared Mathlib tree: **`codelib` (3014 jobs) and `programs` (3024 jobs) build clean, exit 0, no `sorry`/`admit`/`axiom`/`native_decide`** anywhere in the changed `.lean` files. The full `programs` lib (incl. `num_integer`, `total_variation`) is green.

### No correctness defects found
- **Semantics are faithful.** Every chunk theorem matches the interpreter atomic it rests on (`Wp/Atomic.lean`): `i64.div_u`/`i64.rem_u` **trap on divisor 0** — so the `b ≠ 0` precondition is genuinely load-bearing, not decoration; `i64.shl`/`shr_u` mask the amount mod 64; `i64.extend_u` zero-extends u32→u64; `not` = `^ 0xFFFF…FFFF` (`bv_decide`); add/sub/mul are plain wrapping, div/rem unsigned (spot-checked by `#eval`).
- **Calling convention / stack order correct** for every non-commutative op — traced `sub` (`a-b`), `div` (`a/b`), `rem`, `shl`/`shr` (mixed `i32`/`i64` operands) and confirmed they yield `a OP b`, not the reverse. Result expressions match the Rust sources incl. associativity/precedence.
- **Generated modules are clean.** Contiguous `func` lists, no dead/duplicate/gap defs, every export `funcIdx` matches its position, both modules have `imports := []` (so entry index = funcIdx, no offset drift), and all spec entry indices match the export tables. `div`/`rem` bodies match the CodeLib `divBody`/`remBody` verbatim (`call 66`/`67`).
- **The central reuse claim is real and load-bearing**, not cosmetic. `binBodyWp`/`unBodyWp` consume the chunk as an opaque `frag`, so the body theorem reuses the chunk rather than re-deriving it; the inline tests route every opcode through the `…_seq`/`…_chunk` lemma with the raw atomic deliberately excluded from each `simp only`; and the tail-generic `div_chunk`/`rem_chunk` genuinely reuse across the two crates' divergent panic tails (`call 66`/`67` vs `call 74`/`75`).

### Addressed
- **PR description/title were stale** — a snapshot of the superseded 4-operator consolidation: wrong scope (4 of the 11 operators), wrong export indices (`div=3/sub=5/mul=6` vs actual `div=6/sub=8/mul=9`), nonexistent test names (`add_sum3`/`div_chain`/…), and the wrong reuse mechanism (`wp_call_tw`/called-shim instead of the inline `of_wp_entry_for` + chunk-rewrite the code actually uses). **I've rewritten the description** to match `f7a8566`. (No code impact — the proofs and generated tables were always correct; only the prose drifted.)

### Minor / optional (non-blocking, no change required)
- `abs_diff`: the exported `rust_u64::abs_diff` (funcIdx 1) is a trivial `call 0` wrapper; the spec targets the inner helper (funcIdx 0) via `@[spec_of "rust-internal"]`. Sound (the helper is verified and `abs_diff` is symmetric), and pre-existing from #85 — just noting that "abs_diff" coverage is the internal-helper spec by design. Optionally add an export-level spec composing the helper under the `call` rule.
- `entrypoint` (funcIdx 7) is an unspecced call-everything-and-drop driver returning `()`; the gap in the `spec_of` sequence is intentional. A one-line comment would make that obvious.
- The polymorphic `UIntWasm` trunk currently has a single instance (`UInt64`); the generality pays off once a `U32` sibling lands (forward-looking, harmless).

Nice work — the chunk/trunk factoring and the inline-reuse tests are a clean demonstration that one theorem serves both the called export and inlined client code.

<sub>Reviewed with Claude Code (independent rebuild + multi-agent adversarial pass).</sub>

mfornet (approved):

PR #93, opened 2026-06-20
examples: verified bignum predicates + square-witness verifier

jdb19937:
## Why I wrote this

I used to work in ZK proving, in between integrating frameworks I built some demo circuits like a Lucas–Lehmer primality test. The thing that stuck with me from all of that is how much of the difficulty lived in the encoding, in how you decide to constrain and layout the numbers and then carry that decision through every single operation.

So when I came to this repo the first thing I did was read through the examples, and i noticed they all worked at a fixed width of u32. I thought it would be fun to push that range out to not-quite-arbitrary-size integers, which I am calling ℕ† ("N-dagger") in this PR.

ℕ† is the set of naturals you can actually represent in this setup: a number lives in linear memory as a little-endian array of limbs, so the only ceiling on how large it can get is how many limbs the pointer width is able to address, and underneath that ceiling the coverage is total. Proving things over ℕ† means the results hold for every number the machine can hold, which feels like a genuine step toward proving wasm programs over all of ℕ (if that can even be made to make sense).

## What's here

The examples build up as a progression, where each one reuses the encoding and the loop reasoning that the previous one established: is-even, is-a-multiple-of-3, equality, addition, multiplication, and finally a check of whether one number is the square of another.

Everything is proven from scratch and the build is green.

### DIFF
--- a/interpreter/Interpreter/Wasm/Examples/Basic.lean
+++ b/interpreter/Interpreter/Wasm/Examples/Basic.lean
@@ -1,4 +1,5 @@
 import Interpreter.Wasm.Examples.IsEven
+import Interpreter.Wasm.Examples.NatBignum
 import Interpreter.Wasm.Examples.SimpleLoop
 import Interpreter.Wasm.Examples.Factorial
 import Interpreter.Wasm.Examples.InfiniteLoop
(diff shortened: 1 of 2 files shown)


mfornet:
Thanks for the contribution.

I've been going through the PR, and my main concern is that this is just "a take" on how to implement big numbers, but is not clear that is the best way to do it, or that it will be future proof. Notice our goal is not to write WASM manually, but to compile to WASM from other programming langauges (we are focusing on rust for now), and we must make sure our API can talk and reason about artifacts produced by rust.

I'm leaning toward basing this PR on top of any bignum crate. Since there are more than one, we could have a bignum spec, and then prove that each crates satisfies the spec, so we can reuse theorems and lemmas. I'm going to close this for now.

Feel free to reach out in telegram (t.me/TalosDev) to discuss this further, or open an issue.

PR #94, opened 2026-06-22
interpreter: trap on a null array operand in array.copy

jasisz:
`array.copy` with a null source or destination array returns `Invalid` instead of trapping. Per the GC spec a null array operand traps even when `n = 0`.

```wat
(module (type $a (array (mut i32)))
    (array.copy $a $a (array.new_default $a (i32.const 1)) (i32.const 0)
                      (ref.null $a) (i32.const 0) (i32.const 0))
    (i32.const 9)))
```

Before: `Invalid: arrayCopy: ill-shaped operand stack`. After: traps.

The `.arrayCopy` arm matched only non-null operands and fell through to `Invalid`; every sibling array op (`arrayGet` / `arraySet` / `arrayFill` / …) already has the null-trap arm. This adds it for both the source and destination operands.

Tested against **wasmtime 43** and **V8 (Node 26)** as references — both trap with a null-reference error; the patched interpreter now matches. Coverage checked: null at source / destination / both with `n = 0` and `n = 1` → trap; out-of-bounds copy → unchanged; valid copies (`n = 0`, `n = 3`) → unchanged.


mfornet:
Thanks for the contribution 🙌

mfornet (approved):

PR #97, opened 2026-06-22
interpreter: apply active element segments stored in the const-expr form

jasisz:
Active element segments stored as const-expr items (`ref.func`, `ref.null`) get ignored at instantiation — `initialStore` only reads `seg.funcs`, so the table is left full of nulls. The binary format uses this form for typed function tables (`(table N (ref null $ft))`) and GC reference tables, so those come up empty when built from an elem segment.

```wat
(module
  (type $ft (func (result i32)))
  (func $h (type $ft) (i32.const 16))
  (table 1 (ref null $ft))
  (elem (i32.const 0) (ref null $ft) (ref.func $h))
    (call_ref $ft (table.get (i32.const 0)))))
```

- wasmtime 43 → `16`
- V8 (Node 26) → `16`
- Talos before → `Invalid: callRef: ill-shaped operand stack`

`ElementSegment.values` already handles both the `funcs` and `exprs` forms — init just wasn't using it. Moved it (and `evalConstRef`) next to `initialStore`, called it there.

### DIFF
--- a/interpreter/Interpreter/Wasm/Semantics.lean
+++ b/interpreter/Interpreter/Wasm/Semantics.lean
@@ -119,27 +119,6 @@ def readStorageLE (storage : StorageType) (bytes : List UInt8) (off : Nat) : Val
   | .val .f64 => .f64 (UInt64.ofNat n)
   | .val _    => .i32 0
 
-/-- Evaluate a simple element-segment item constant expression to its
-reference value (GC proposal). Covers the forms element segments use:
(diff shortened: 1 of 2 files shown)


jasisz:
Closing this — my repro instantiated the module without running the const-expr element initializers (`runConstElems`), so the table looked empty. On a full instantiation it's populated correctly; there's no bug here. Apologies for the noise.

PR #98, opened 2026-06-22
u64: comparisons (eq..ge) + Ord (min/max/clamp)

El3ssar:
Continues the Rust-std → Wasm → Lean u64 corpus with the **comparisons + Ord** block. All proofs green, 0 `sorry`, 0 warnings.

## What's here (7 commits, one per function except the cohesive Ord block)

**Comparisons — new `cmp` shape (`eq ne lt le gt ge`).** Adds `CmpChunk`/`cmpBodyWp` to the type-agnostic trunk `CodeLib/RustStd/UInt.lean`: a comparison chunk whose result lands as the i32 boolean `0/1` (not a `toV`-carried `T`). A comparison `a OP b` compiles to `[cmpOp, i32.const 1, i32.and]`; the redundant `& 1` bool-mask is discharged inside each `<fn>_chunk`. `U64/Eq.lean` is the exemplar; `Ne/Lt/Le/Gt/Ge` mirror it (opcode + `res`). Each has the export-body theorem + an inline `_seq` restatement, with two structurally-distinct inline reuse tests in `rust_u64_tests` (the op's own `wp_*` atomic excluded from `simp`, delete-break-verified load-bearing).

**Ord — framed `select` shape (`min max clamp`).** These compile to wrapper exports that `call` framed inner functions (the abs_diff pattern), reused across the `call` boundary via `wp_call_tw`:
- `CodeLib/RustStd/U64/{Max,Min,Clamp}.lean`: body theorems `max_wp`/`min_wp`/`clamp_wp`. `clamp` panics if `lo > hi` (spec carries `lo ≤ hi`), saves/restores the global stack pointer, and is **tail-generic in the panic `call` index** so one theorem serves every crate's copy.
- `Frame.lean`: new `Mem.read64_write64_disjoint` (+ `write64_bytes_outside`) for the two-operand spill frames (min/max read back from adjacent 8-byte slots).
- Per-crate `MaxSpec`/`MinSpec`/`ClampSpec` prove the exported wrappers by reusing the inner body theorems via the call rule. Two call-reuse tests each in `rust_u64_tests` (`max_add`/`max_chain`, `min_add`/`min_chain`, `clamp_add`/`clamp_mul`).

## Notes for review
- Adding the Ord wrappers **renumbers the whole module** (`opt-0` inserts the framed inner fns at low indices): all merged + comparison spec indices were re-mapped, and the div/rem panic-tail `.call`/`.const` literals updated (72/73 → 83/84). The mechanical re-index is the bulk of the `RustU64/Spec.lean` / `Program.lean` diff.
- `clamp_wp` needs `maxHeartbeats 1000000` (~42s) — the nested-block `simp`; functionally fine, optimizable later.
- The `cmp` trunk shape and the operator trunk are polymorphic, so u32/u16/u8/signed reuse them (swap opcode + `res`); the framed `select` family will re-prove per width on `Frame.lean`.

### DIFF
--- a/codelib/CodeLib.lean
+++ b/codelib/CodeLib.lean
@@ -20,6 +20,7 @@ import CodeLib.RustStd.U64.Shl
 import CodeLib.RustStd.U64.Shr
 import CodeLib.RustStd.U64.Eq
 import CodeLib.RustStd.U64.Ne
+import CodeLib.RustStd.U64.Lt
 import CodeLib.RustStd.Option
 import CodeLib.Near.State
 import CodeLib.Near.Env
--- a/codelib/CodeLib.lean
+++ b/codelib/CodeLib.lean
@@ -23,6 +23,7 @@ import CodeLib.RustStd.U64.Ne
 import CodeLib.RustStd.U64.Lt
 import CodeLib.RustStd.U64.Le
 import CodeLib.RustStd.U64.Gt
+import CodeLib.RustStd.U64.Ge
 import CodeLib.RustStd.Option
 import CodeLib.Near.State
 import CodeLib.Near.Env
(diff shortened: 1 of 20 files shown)


mfornet:
Talked to @El3ssar off-line, we are closing this in favor of having a slightly more general version of BinChunk and re-using it here.

PR #100, opened 2026-06-22
interpreter: read table.init values from the segment, incl. const-expr items

jasisz:
`table.init` copied from the per-store `elementSegments` (a func-index list), so a passive element segment in the const-expr form (`ref.func`, …) — whose func-index list is empty — copied nothing, leaving the table null.

```wat
(module
  (type $ft (func (result i32)))
  (func $h (type $ft) (i32.const 16))
  (table 1 (ref null $ft))
  (elem $e (ref null $ft) (ref.func $h))
    (table.init $e (i32.const 0) (i32.const 0) (i32.const 1))
    (call_ref $ft (table.get (i32.const 0)))))
```

- wasmtime 43 → `16`
- V8 (Node 26) → `16`
- Talos before → traps (`table.get` returns the null `table.init` left).

`array.new_elem` already reads the segment's values via `ElementSegment.values` (which handles both forms); `table.init` now does the same, keeping `elementSegments` only for the dropped check. Same gap as #97, on the passive side.


jasisz:
This PR makes `table.init` read the segment's values, so `ref.func` / `ref.i31` items land correctly. What it does **not** fix — and I'm not touching it here, just putting it on the record — is the rest of the const-expr items.

The values path (`ElementSegment.values` / `evalConstRef`) only evaluates `ref.func` / `ref.null` / `ref.i31` / `const`. Anything else that's valid — `struct.new`, `global.get`, `array.new_default` — falls through to a null default, so those items silently become null instead of the real reference.

So even with this PR merged, a struct item still comes out null:

```wat
(module (type $s (struct (field i32))) (type $arr (array (mut anyref)))
  (elem $e anyref (item (struct.new $s (i32.const 42))))
    (array.new_elem $arr $e (i32.const 0) (i32.const 1)) (array.get $arr (i32.const 0)) ref.is_null))
```
wasmtime / V8 → `0` (real struct); Talos → `1` (null). (`array.new_elem` already had this, and `evalConstRef`'s docstring even says `struct.new` isn't handled.)

This needs a separate fix that I don't want to add here.

mfornet (approved):

PR #104, opened 2026-06-23
u64: generalize reusable integer chunks

El3ssar:
## Summary
- Generalize Rust integer chunk helpers to read operands from arbitrary locals, including heterogeneous and side-conditioned binary chunks.
- Refactor U64 div/rem/shift helpers to avoid hardcoded exported bodies and panic tails in CodeLib.
- Update RustU64 specs and reuse tests to compose the new chunk helpers.

## Test plan
- `cd codelib && lake build`
- `cd programs/lean && lake build`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/codelib/CodeLib/RustStd/U64/Add.lean
+++ b/codelib/CodeLib/RustStd/U64/Add.lean
@@ -1,23 +1,28 @@
 import CodeLib.RustStd.U64.Basic
 
-/-! `u64::add` — inlined to a single `i64.addI64`. Chunk fact + export body, reusing the trunk. -/
+/-! `u64::add` — inlined to local reads plus a single `i64.addI64`. Chunk fact + body, reusing the trunk. -/
 
 namespace Wasm.RustStd.U64
 open Wasm Wasm.RustStd
 
-/-- The inlined chunk `[.addI64]` computes `+` on stack operands (reusable wherever
-`a + b` is inlined). -/
+/-- The inlined chunk `[.addI64]` computes `+` after operands are read from locals. -/
 theorem add_chunk : BinChunk (T := UInt64) [.addI64] (· + ·) := by
-  intro α m env Q st P L rest a b vs
-  simp only [List.cons_append, List.nil_append, toV_u64, wp_addI64_cons]
+  intro α m env Q st P L rest i j a b vs ha hb
+  simp only [toV_u64] at ha hb
+  have hb' : (⟨P, L, .i64 a :: vs⟩ : Locals).get j = some (.i64 b) := by
+    simpa [Locals.get] using hb
+  simp only [List.cons_append, List.nil_append, toV_u64, wp_localGet_cons, ha, hb',
+    wp_addI64_cons]
 
 set_option maxRecDepth 4096 in
-/-- Export-body theorem for `rust_u64::add`, built by reusing `add_chunk`. -/
+/-- Fallthrough body theorem for `rust_u64::add`, built by reusing `add_chunk`. -/
 theorem addBodyWp {α} {m : Module} {env : HostEnv α} (st : Store α)
     (a b : UInt64) (vs : List Value) :
-    wp m ([.localGet 0, .localGet 1] ++ [.addI64] ++ [.ret])
-      (Returns (.i64 (a + b) :: vs) (framePost st)) st ⟨[.i64 a, .i64 b], [], vs⟩ env :=
-  binBodyWp add_chunk st a b vs
+    wp m ([.localGet 0, .localGet 1] ++ [.addI64])
+      (fun c => ∃ st',
+        c = .Fallthrough st' ⟨[.i64 a, .i64 b], [], .i64 (a + b) :: vs⟩ ∧ framePost st st')
+      st ⟨[.i64 a, .i64 b], [], vs⟩ env :=
+  binBodyWp add_chunk st 0 1 a b vs rfl rfl
 
 /-- Concrete `i64` restatement of `add_chunk` (matches emitted opcodes for `rw`
 at an inlined `i64.add`), proven *by reusing* the polymorphic `add_chunk`. -/
@@ -25,6 +30,6 @@ theorem add_seq {α : Type} {m : Module} {env : HostEnv α} {Q : Assertion α}
     {st : Store α} {P L : List Value} {rest : Program} (a b : UInt64) (vs : List Value) :
     wp m (.addI64 :: rest) Q st ⟨P, L, .i64 b :: .i64 a :: vs⟩ env ↔
       wp m rest Q st ⟨P, L, .i64 (a + b) :: vs⟩ env :=
-  add_chunk a b vs
+  by simp only [wp_addI64_cons]
 
 end Wasm.RustStd.U64
--- a/codelib/CodeLib/RustStd/U64/Add.lean
+++ b/codelib/CodeLib/RustStd/U64/Add.lean
@@ -1,35 +1,23 @@
 import CodeLib.RustStd.U64.Basic
 
-/-! `u64::add` — inlined to local reads plus a single `i64.addI64`. Chunk fact + body, reusing the trunk. -/
+/-! `u64::add` — inlined to a single `i64.addI64`. Chunk fact + concrete
+restatement, reusing the trunk. -/
 
 namespace Wasm.RustStd.U64
 open Wasm Wasm.RustStd
 
-/-- The inlined chunk `[.addI64]` computes `+` after operands are read from locals. -/
-theorem add_chunk : BinChunk (T := UInt64) [.addI64] (· + ·) := by
-  intro α m env Q st P L rest i j a b vs ha hb
-  simp only [toV_u64] at ha hb
-  have hb' : (⟨P, L, .i64 a :: vs⟩ : Locals).get j = some (.i64 b) := by
-    simpa [Locals.get] using hb
-  simp only [List.cons_append, List.nil_append, toV_u64, wp_localGet_cons, ha, hb',
-    wp_addI64_cons]
+/-- The reusable chunk: `[.addI64]` computes `+` on stack operands. Feeds the
+export-body spec via `binBodyReturnsWp`; also `rw`-able at an inlined `i64.add`. -/
+theorem add_chunk : BinChunk [.addI64] ((· + ·) : UInt64 → UInt64 → UInt64) := by
+  intro α m env Q st P L rest a b vs _
+  simp only [List.cons_append, List.nil_append, toV_u64, wp_addI64_cons]
 
-set_option maxRecDepth 4096 in
-/-- Fallthrough body theorem for `rust_u64::add`, built by reusing `add_chunk`. -/
-theorem addBodyWp {α} {m : Module} {env : HostEnv α} (st : Store α)
-    (a b : UInt64) (vs : List Value) :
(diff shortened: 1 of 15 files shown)


mfornet (commented):
Automated code review (medium effort). No correctness bugs found — operand order on the non-commutative ops, the div/rem zero-guard, and the shift mask are all preserved, and `lake build` passes on both `codelib` and `programs/lean`. The comments below are all cleanup-class: the recurring theme is that this PR introduced several reusable helpers (`nonzeroGuardWp`, the `*Where` body helpers, the bv_decide identities) but didn't actually consume them, leaving dead code and duplicated expensive proofs.

codelib/CodeLib/RustStd/UInt.lean
```diff
@@ -6,89 +6,216 @@ import CodeLib.Entry
...
+          c = .Fallthrough st' ⟨P, L, toV (op a b) :: vs⟩ ∧ framePost st st')
+        st ⟨P, L, vs⟩ env := by
+    rw [chunk i j a b vs ha hb]
+    unfold framePost
+    simp
+  simpa only [List.append_nil] using h
+
+/-- Turn a side-conditioned heterogeneous binary chunk into a fallthrough theorem
+for any local frame. -/
+theorem hbinBodyWpWhere {A B C : Type} [UIntWasm A] [UIntWasm B] [UIntWasm C]
```
mfornet: **Dead code.** `hbinBodyWpWhere` (here) and `hbinBodyWhereReturnsWp` (~line 151) are added by this PR but have zero call sites — `hbinBodyWhereReturnsWp` is referenced nowhere, `hbinBodyWpWhere` only in a doc comment. Div/Rem (the only side-conditioned ops) bypass them and hand-roll `divBodyWp`/`remBodyWp`. Either delete both, or route Div/Rem through them so the generic guarded-body path is the one that's actually exercised.

codelib/CodeLib/RustStd/U64/Add.lean
```diff
@@ -1,30 +1,35 @@
...
+  simp only [toV_u64] at ha hb
+  have hb' : (⟨P, L, .i64 a :: vs⟩ : Locals).get j = some (.i64 b) := by
+    simpa [Locals.get] using hb
+  simp only [List.cons_append, List.nil_append, toV_u64, wp_localGet_cons, ha, hb',
+    wp_addI64_cons]
 
 set_option maxRecDepth 4096 in
-/-- Export-body theorem for `rust_u64::add`, built by reusing `add_chunk`. -/
+/-- Fallthrough body theorem for `rust_u64::add`, built by reusing `add_chunk`. -/
 theorem addBodyWp {α} {m : Module} {env : HostEnv α} (st : Store α)
```
mfornet: **Dead code (applies to all 9 per-op `*BodyWp`).** `addBodyWp` and its siblings (`subBodyWp`, `mulBodyWp`, `bitand/bitor/bitxorBodyWp`, `notBodyWp`, `shlBodyWp`, `shrBodyWp`) are referenced only in doc prose (`RustU64/Spec.lean:7`, `:354`). Every export in `RustU64/Spec.lean` is discharged directly via `binBodyReturnsWp`/`hbinBodyReturnsWp`/`unBodyReturnsWp` applied to the chunk, never through these. They're maintained and re-verified every CI run while protecting nothing, and the obvious name (`addBodyWp`) misdirects readers away from the live `binBodyReturnsWp`. Delete, or have Spec consume them.

programs/lean/Project/RustU64Tests/Spec.lean
```diff
@@ -275,7 +275,7 @@ theorem div_then_add_correct : DivThenAddSpec := by
     List.getElem?_cons_zero, List.getElem?_cons_succ, Nat.reduceAdd, Nat.reduceLT, reduceIte,
     List.drop, wp_constI64_cons, wp_eqI64_cons, hb, ↓reduceIte, wp_const_cons, wp_and_cons,
     wp_br_if_cons, h10]
```
mfornet: **New lemma not reused.** This PR extracts the zero-divisor guard into `nonzeroGuardWp` (`Basic.lean:38`, used in `Div.lean:49`/`Rem.lean:50`), but the four div/rem test proofs (here and ~298, ~322, ~345) still inline the guard by hand: `have h10 : (1:UInt32) &&& 0 = 0 := by decide` + this ~17-lemma simp set. The guard's emitted opcode shape is now encoded in two unsynchronized places; a change to how opt-0 emits the divisor check forces editing `nonzeroGuardWp` *and* four copy-pasted `h10` blocks. Replace the four with `rw [nonzeroGuardWp ...]`.

codelib/CodeLib/RustStd/U64/Shl.lean
```diff
@@ -1,46 +1,46 @@
...
+theorem shl_chunk :
+    HBinChunk (A := UInt64) (B := UInt32) (C := UInt64)
+      (shiftAmountFrag ++ [.shlI64])
+      (fun a b => a <<< (b.toUInt64 % 64)) := by
+  intro α m env Q st P L rest i j a b vs ha hb
+  simp only [toV_u64, toV_u32] at ha hb
+  have hb' : (⟨P, L, .i64 a :: vs⟩ : Locals).get j = some (.i32 b) := by
+    simpa [Locals.get] using hb
+  have hamt : a <<< (UInt64.ofNat (shiftMask &&& b).toNat % 64) = a <<< (b.toUInt64 % 64) := by
+    simp; bv_decide
```
mfornet: **Duplicated `bv_decide`.** `shl_chunk` (here) and `shl_seq` (line 31) each run `simp; bv_decide` independently on the same shift-amount masking identity; same duplication in `Shr.lean` (19/31) and `Not.lean` (17/38) for the complement identity. Pre-PR the `_seq` lemma reused the chunk; now the chunk requires a `localGet` prefix so `_seq` re-proves from scratch, raising the shift `bv_decide` count from 2 to 4. `bv_decide` calls an external SAT solver and is among the slowest tactics in the build — factor each identity into one shared lemma proven once and reuse it in both.

codelib/CodeLib/RustStd/U64/Add.lean
```diff
@@ -1,30 +1,35 @@
...
 
-/-- The inlined chunk `[.addI64]` computes `+` on stack operands (reusable wherever
-`a + b` is inlined). -/
+/-- The inlined chunk `[.addI64]` computes `+` after operands are read from locals. -/
 theorem add_chunk : BinChunk (T := UInt64) [.addI64] (· + ·) := by
-  intro α m env Q st P L rest a b vs
-  simp only [List.cons_append, List.nil_append, toV_u64, wp_addI64_cons]
+  intro α m env Q st P L rest i j a b vs ha hb
+  simp only [toV_u64] at ha hb
+  have hb' : (⟨P, L, .i64 a :: vs⟩ : Locals).get j = some (.i64 b) := by
```
mfornet: **Copy-paste across 10 files.** This `have hb' : (⟨P, L, .i64 a :: vs⟩ : Locals).get j = some (...) := by simpa [Locals.get] using hb` re-derivation is repeated verbatim in all ten chunk proofs (Add/Sub/Mul/BitAnd/BitOr/BitXor/Div/Rem/Shl/Shr). It re-states `hb` after one `localGet` pushed a value, because `Locals.get` ignores the value stack — a generic fact. Lift it to one lemma next to `Locals.get` (e.g. `Locals.get_values_irrel`), or make `wp_localGet_cons` discharge the second read, so a refactor of `Locals.get` is a 1-place edit instead of 10.

codelib/CodeLib/RustStd/U64/Div.lean
```diff
@@ -1,40 +1,55 @@
...
     wp m (.divUI64 :: rest) Q st ⟨P, L, .i64 b :: .i64 a :: vs⟩ env ↔
       wp m rest Q st ⟨P, L, .i64 (a / b) :: vs⟩ env := by
   simp only [wp_divUI64_cons, hb, ↓reduceIte]
 
-/-- Verbatim opt-0 body of `rust_u64::div`. -/
-def divBody : Program :=
-  [ .block 0 0 [ .localGet 1, .constI64 0, .eqI64, .const 1, .and, .br_if 0,
-                 .localGet 0, .localGet 1, .divUI64, .ret ],
-    .const 1048600, .call 66, .unreachable ]
+/-- Guarded divide body fragment, parameterized by dividend/divisor locals. The
```
mfornet: **Duplicated guarded-body proof.** `divCheckedBody`/`divBodyWp` here and `remCheckedBody`/`remBodyWp` in `Rem.lean` are byte-identical except for the `divUI64`/`remUI64` opcode, including the fragile `apply wp_block_cons; change …; rw [List.append_assoc …]; rw [nonzeroGuardWp …]; change …; rw [chunk …]; simp` ladder. A single `checkedBinBody` combinator in `UInt.lean` (parameterized over the chunk) would absorb both and give the next trapping op (signed div, checked shift) a real template instead of a third paste.

codelib/CodeLib/RustStd/U64/Add.lean
```diff
@@ -1,30 +1,35 @@
...
-/-! `u64::add` — inlined to a single `i64.addI64`. Chunk fact + export body, reusing the trunk. -/
+/-! `u64::add` — inlined to local reads plus a single `i64.addI64`. Chunk fact + body, reusing the trunk. -/
 
 namespace Wasm.RustStd.U64
 open Wasm Wasm.RustStd
 
-/-- The inlined chunk `[.addI64]` computes `+` on stack operands (reusable wherever
-`a + b` is inlined). -/
+/-- The inlined chunk `[.addI64]` computes `+` after operands are read from locals. -/
 theorem add_chunk : BinChunk (T := UInt64) [.addI64] (· + ·) := by
```
mfornet: **Boilerplate the trunk could absorb.** The six homogeneous `BinChunk` proofs (`add`/`sub`/`mul`/`bitand`/`bitor`/`bitxor_chunk`) are identical except for the operator and the trailing atomic lemma (`wp_addI64_cons` etc.). A generic `binChunk_of_atomic` in `UInt.lean` keyed on the atomic `wp_*` lemma would collapse each to a one-line application — matching this PR's stated goal of proving reusable shapes once in the trunk. As-is, any change to how locals are read is a 6-place edit.

programs/lean/Project/RustU64/Spec.lean
```diff
@@ -70,7 +70,8 @@ def DivSpec : Prop :=
 theorem div_correct : DivSpec := by
   intro env a b hb
   exact (TerminatesWith.of_returns_wp (f := func6Def) (rs := [.i64 (a / b)]) rfl rfl
-      (divBodyWp «module».initialStore a b [] hb) rfl).mono (fun _ _ h => h.1)
+      (divBodyWp «module».initialStore 0 1 a b [] [.const 1048600, .call 66, .unreachable]
```
mfornet: **Magic literals moved to the call site.** The panic tail `[.const 1048600, .call 66, .unreachable]` (and `1048616`/`.call 67` for rem) is passed inline here. These data-offsets and imported-func indices were previously encapsulated in `divBody`/`remBody`; the reusable `divCheckedBody` fragment excludes the tail, so the encapsulation moved to the call site rather than being removed. If the wasm is regenerated and the offset or func index shifts (the opt-level-0 churn this project has hit before), a wrong value still typechecks the fragment and only fails deep in the `rfl` discharge.

mfornet (approved):

PR #105, opened 2026-06-23
Add reusable Rust slice len/is_empty proofs

El3ssar:
## Summary
- Add generic Rust slice primitive corpus for `&[T]` `len` and `is_empty`, with concrete wasm monomorphizations for verifier emission.
- Add reusable CodeLib WP lemmas for slice length and emptiness over the wasm fat-pointer length component.
- Add `rust_array_tests` reuse proofs with two non-identical test functions per primitive.

## Test plan
- `lake build CodeLib CodeLib.RustStd.Array.Len CodeLib.RustStd.Array.IsEmpty`
- `just verifier-prove rust_array rust_array_tests`
- Lean verify on representative codelib/test theorems

Made with [Cursor](https://cursor.com)


El3ssar (commented):
**Automated adversarial review** (independent; no prior context on this change). I read every changed file in full, the interpreter WP lemmas the proofs rest on (`wp_localGet_cons`, `wp_load32_cons`, `Returns`/`of_returns_wp`/`of_wp_entry_for`/`wp_call_tw`, the `Mem`/`read32` model), the generated `Program.lean` against the checked-in `program.wat`, and the Rust sources. I also re-checked every new Lean file with `lake env lean` — **all five new/changed Lean files typecheck clean, no `sorry`/`axiom`/`native_decide`**, and both `Spec.lean` files verify fresh. So the headline is: the proofs are *sound and complete*; my findings are about over-claiming, dead surface area, faithfulness nuance, and a cross-crate inconsistency — not broken proofs.

## What's actually correct (verified, stated once)

- **Faithfulness to emitted wasm:** `RustArray/Program.lean` and `RustArrayTests/Program.lean` match `programs/rust/build/*/program.wat` instruction-for-instruction (func indices, params, loads at offset 0/4, call targets, the `& 1` re-masks). Operand order, local indices, and the `(ptr,len)` fat-pointer layout in the proofs all line up with the `.wat`.
- **Export memory specs are non-vacuous and correctly bounded.** `LenExportSpec`/`IsEmptyExportSpec` precondition `read32 (p+0)=dataPtr`, `read32 (p+4)=len`, `p.toNat + 8 ≤ pages*65536`. The bound is exactly tight for the two `load32`s (offsets 0 and 4 both need +4); `wp_load32_cons` checks `a.toNat+off.toNat+4 > pages*65536`, and `p.toNat+8 ≤ …` implies both. The memory hypotheses are satisfiable for any store (pick `dataPtr`/`len` = the actual reads), so nothing is vacuous or contradictory. `fatPtrLoadWp` is genuinely polymorphic over `Q/rest/vs/P/L` — the "reusable loader" claim holds.
- **`LenChunk`/`lenOpBodyWp` are honestly reusable**: quantified over store, `vs`, locals, `Q`, and `rest`; nothing load-bearing is silently pinned.

## Findings

### 1. `isEmpty_seq` is dead code (major — minimal-surface-area)
`codelib/CodeLib/RustStd/Array/IsEmpty.lean:44-49`. `isEmpty_seq` (the "inlined" half of the dual-shape API) has **zero consumers** — grep shows references only inside its own docstring. Its own docstring concedes "At opt-0 the compiler emits `is_empty` as a *call*," and indeed both test users (`empty_plus_three`, `empty_xor_flag`) and both export wrappers go through `.call`, never the inline fragment. The `.wat` confirms `core::slice::is_empty` is never inlined. So this is speculative future-use surface area with no exercised path, in a corpus that prizes minimal surface area. **Fix:** drop `isEmpty_seq` (and its half of the "dual-shape" prose) until a consumer needs the inline form, or add a hand-written inlined example that exercises it.

### 2. The `len` side of the chunk abstraction is content-free (minor)
`Len.lean:15` `len_chunk : LenChunk [] id` and `len_seq` (`Len.lean:23`) are degenerate: `len_seq` is literally `wp_localGet_cons` specialized to `.i32 len` (the docstring admits "the reusable content is exactly `wp_localGet_cons`"), and `len_chunk` has `frag=[]`, `op=id`, so it carries no computational content. `lenBodyWp` therefore reduces `[localGet i, .ret]` to "return the local," which `wp_localGet_cons`+`wp_ret_cons` already give. This is defensible as *architectural symmetry* with `is_empty` (one `LenChunk` template for both), and `lenBodyWp` is at least exercised by `len_call`/`len_correct`. But contrast the sibling `RustU64Tests`, whose methodology deliberately removes the atomic `wp_*` from the simp set so "the reusable CodeLib theorem is the only way through (confirm by dropping the chunk lemma: the proof then fails)." For `len`, dropping the chunk changes nothing. **Fix:** either keep `len` purely as the symmetry demonstration and say so plainly in the docstring (drop the "reusable chunk" framing for the `[]/id` case), or fold `len_seq` away (callers can use `wp_localGet_cons` + the `get` fact directly).

### 3. Cross-crate `spec_of` kind inconsistency for the *tests* crate (major)
`RustArrayTests/Spec.lean:23,44,69,90` tag all four specs `"rust-internal"`, targeting the internal monomorphized bodies (`«module» 1/0/4/2`) with the raw `[len, ptr]`/`[…, len, ptr]` calling convention. The directly-analogous sibling `RustU64Tests/Spec.lean` tags the *same situation* (internal monomorphized body, named after an exported symbol) `"rust-exported"`. So the two "tests" crates classify identical constructs under different registry kinds. Independently of which is "right," they must agree. Note the names (`rust_array_tests::len_plus_one`, …) are the names of `#[no_mangle] pub extern "C"` **exports**, but these specs cover the inner `(len,ptr)` body, not the export ABI. **Fix:** pick one convention for tests-crate internal-body specs and apply it to both crates; if `rust-internal` is intended, reconcile `RustU64Tests`; either way, document what `rust-internal` vs `rust-exported` means for a tests crate.

### 4. The test crate's actual exports (func5–func8) are entirely unverified (minor)
`RustArrayTests` exports `empty_plus_three`/`empty_xor_flag`/`len_plus_arg`/`len_plus_one` as func5–func8 (fat-pointer-in-memory ABI), but every spec stops at the internal body (func0/1/2/4). No theorem covers the four exported entry points. The primitive `RustArray` crate *does* prove its exports (`LenExportSpec`/`IsEmptyExportSpec`), so the asymmetry is conspicuous. This is acceptable *if* the stated purpose is only to demonstrate chunk reuse on varied internal body shapes — but then the `spec_of` names pointing at the export symbols (finding #3) overstate coverage. **Fix:** either add export specs for func5–func8 (reusing `fatPtrLoadWp` like `RustArray`), or rename/annotate the test specs so they don't claim the exported symbol.

### 5. Docstring overstates inlining for the test crate (nit)
`rust_array_tests/src/exports.rs:4-5` and `RustArrayTests/Spec.lean` prose say the tests reuse chunks "against the emitted **inline** instruction sequences." True for `len_plus_one`/`len_plus_arg` (length read is `local.get 1`), but **false for `is_empty`**: `empty_plus_three`/`empty_xor_flag` reach `is_empty` via `call $core::slice::is_empty` (see `program.wat:50,32`), exercising the *called* `isEmptyBodyWp`, not an inline fragment. **Fix:** say "inline for `len`, called for `is_empty` at opt-0."

### 6. Trap/OOB behavior of the exports is unspecified (nit)
`LenExportSpec`/`IsEmptyExportSpec` are total correctness *conditional on* `p.toNat+8 ≤ pages*65536`. Nothing asserts that an out-of-bounds `p` traps (the `wp_load32_cons` trap branch is never reached). That's the standard partial-domain spec shape and fine, but worth one line in the docstring so a reader doesn't assume OOB safety is covered.

### Naming nuance (informational, not a defect)
`func0` in `RustArray` is `rust_array::len` (the user's `len<T>` wrapper), which the compiler reduced to `local.get 1; return` because `core::slice::len` was inlined — so the Basic.lean prose about "`&[T]::len` … the monomorphized primitive body just returns it" is accurate in effect, just note it's the *user* wrapper post-inline, not `core::slice::len` as a distinct function. No action needed.

## Verdict
**Proofs are correct and the wasm modeling is faithful** — I could not find an unsound, vacuous, or mis-stated *theorem*, and the memory-resident export specs are right. The issues are quality/over-claiming: one dead lemma (`isEmpty_seq`), a content-free `len` chunk dressed as substantive reuse, a cross-crate `spec_of` kind inconsistency, unverified test-crate exports under export-named specs, and two docstring overstatements. None block correctness; #1 and #3 should be resolved before merge for a corpus that values minimal, honestly-labeled surface area. The PR is still in **Draft**, which is the right state. Recommend: trim/rename, reconcile the tag convention with `RustU64Tests`, and tighten the prose.

El3ssar:
## Review addressed — pushed in `21d8135`

**`isEmpty_seq` dead code (major).** Removed. The inline-reuse capability is not lost — it lives in `isEmpty_chunk`, the canonical stack-form unit (which already feeds `isEmptyBodyWp`); a future inlined `is_empty` `rw`s with the chunk directly. Docstrings updated to say so.

**`spec_of` tag inconsistency + unverified test exports (major + minor).** Fixed together: the `rust_array_tests` exported wrappers `func5`–`func8` are now verified end-to-end through the fat-pointer memory marshalling — `fatPtrLoadWp` loads `(dataPtr, len)` back, then per-impl call bridges (which reuse `len_seq` / `isEmptyBodyWp`) discharge the body — and tagged `@[spec_of "rust-exported" …]`, matching the `rust_u64_tests` convention. The crate now carries both the `rust-internal` inlined-reuse specs and the `rust-exported` end-to-end specs, so the exported symbols are no longer named but unproven.

**`len` chunk content-free (minor).** Kept (`len_seq` is `rw`-ed by the two `len` reuse tests) but the docstrings are now honest: `len_chunk : LenChunk [] id` is the trivial (`op = id`) case, definitionally `wp_localGet_cons`, and `len`'s load-bearing reuse is the called body `lenBodyWp` — not the inline form.

**`exports.rs` overstates inlining (nit).** Fixed the Rust doc comments: at opt-0 `len` inlines to a length read while `is_empty` lowers to a `call`, so both shapes the corpus targets are exercised (and that is what the proofs reuse).

**Export OOB/trap unspecified (nit).** Documented: the export specs are *conditional total correctness* — given the in-bounds ABI contract the call terminates with the right value; the out-of-bounds `load32` trap is outside the contract and deliberately not asserted.

mfornet (approved):

Issue #68, opened 2026-06-15
Memory for proof & spec maintainability


Isabelle9999:
**Problem:** every time a proof writes to address A and reads from address B, it needs a lemma saying "if A ≠ B, the read returns the old value." Currently each proof writes its own memory disjointness lemmas from scratch instead of reusing generic ones (e.g itoa has ~10 such lemmas) and every new program re-derives them.

**Solution:** add region-based framing lemmas in codelib ( no interpreter changes)

Define a MemRegion type ( start address + length) and provide generic lemmas
- read_write_same: if addr is within the written region, read returns the new value
- read_write_disjoint: if the read region and write region don't overlap, read returns the old value
- write_write_commute: writes to disjoint regions commute
This helps with maintainability of proofs.

For spec readability (mentioned previously), add a MemArray type on top of MemRegion that wraps a base pointer + element size + count. Specs would use _arr.get i_ instead of _st.mem.read32(ptr + 4 * i)_, more readable.

Possible automating it: the verifier already reads the Wasm binary to generate Program.lean,  it could also generate framing lemmas automatically by analyzing which memory ranges each function touches.

Reference:  CompCert Memory Model (Leroy & Blazy, "Formal Verification of a C-like Memory Model," 2008; Leroy & Appel, "A Formally Verified Compiler Back-end," 2012).

theebayuser:
I took a first cut at this: https://github.com/cajal-technologies/talos/pull/131 lifts the six read/write disjointness lemmas that were private to the NumInteger spec into `CodeLib.RustStd.Frame` (with NumInteger refactored to consume them, per the use-site rule).

Suggested phasing for the rest of the issue, to keep each piece attached to a real consumer:

1. **(the PR above)** raw `Mem.read{32,64}_write{32,64}_disjoint` framing lemmas — the `read_write_same` / `read_write_disjoint` core of this issue.
2. **`write_write_commute` + `MemRegion`** (start + length, region-level disjointness): I'd like to drive these out against the `SwapElementsSpec` proof (#110's spec is statement-only) and land them together with it — swap's "every element `k ≠ i, j` is unchanged" postcondition is exactly a region-framing statement, so it's a good forcing function for getting the region API right.
3. **`MemArray`** (base + element size + count, `arr.get i` spec view) after that, once merge_sort-style targets (#106) give it a consumer.

Happy to adjust scope if you'd rather see it sliced differently.

theebayuser:
[...]

PR #106, opened 2026-06-24
merge_sort: memory-based verification target with deferred spec

mfornet:
## Overview

Adds a new `merge_sort` crate + Lean verification module as a vehicle for exploring **programs that operate on linear memory**. The behavioural spec is written and typechecks; the proof is deliberately deferred.

## What changed

**Rust — `programs/rust/merge_sort/`** (builds clean for `wasm32-unknown-unknown`, clippy-clean)
- `src/lib.rs` — safe, **no `unsafe`**:
  - `pub fn merge_sort(Vec<u32>) -> Vec<u32>` — owned-`Vec` API, delegates to…
  - `sort_slice(&mut [u32])` — allocates the scratch buffer internally and runs…
  - `merge_sort_into(data, scratch)` — **top-down recursive** merge sort over two equal-length slices, plus a stable `merge`.
- `src/exports.rs` — the **only** `unsafe`: the wasm export `merge_sort(data_ptr, len)` reconstructs the data slice from raw parts and calls `sort_slice`.

**Lean — `programs/lean/Project/MergeSort/`**
- `Program.lean` — auto-emitted decode of the wasm module; export at **`funcIdx 33`**.
- `Spec.lean` — total-correctness statement via `TerminatesWith`: the data region ends **sorted ascending** (`List.Pairwise (· ≤ ·)`) and a **permutation** of its input (`.Perm`), with a `wordsAt` helper giving the `[u32]`-as-list view of memory.
- Registered in `Project.lean`.

## Design notes for reviewers

- **Internal allocation is intentional.** An earlier iteration passed `scratch_ptr` across the ABI (allocation-free, trivial memory frame). This version drops it so scratch is allocated *inside* the call — pulling the Rust allocator and `memory.grow` into the verified call graph (module grew 67 → 94 functions). This is deliberate: it's part of what we want to be able to verify going forward.
- **The spec's preconditions got uglier, honestly so.** Because scratch is now allocated internally, the statement pins the allocator's control state to pristine (agreement with `initialStore` below `heapBase`) and adds a **provisional** scratch-disjointness margin (`heapBase + 4*len ≤ data_ptr`). The exact dlmalloc margin — plus proving the allocator terminates, that any `memory.grow` doesn't trap, and that scratch is disjoint from the data — is the deferred allocator-verification work, flagged in the docstring.
- **`List.Pairwise`, not `List.Sorted`** — the pinned Mathlib removed the relation-based `List.Sorted`; `Pairwise (· ≤ ·)` is the idiom for ascending order.
- **No `sorry`.** Per project convention, only the spec `def` is committed (no `@[proves]` theorem). `lake build Project.MergeSort.Spec` succeeds.

## Caveat

`Program.lean` was generated from a **local (non-canonical) wasm build**. Function order — and therefore the export `funcIdx` (33) referenced in the spec — can shift under the pinned canonical Docker build. Regenerate canonically before starting the proof.


theebayuser:
Noticed `Project/MergeSort/Framing.lean` here reinvents a chunk of memory-region algebra that either already landed in `CodeLib` or is one 32-bit twin away in the stack I have open for #68 — flagging the overlap in case it lets this file shrink to imports.

**Already in codelib.** The docstring notes "what is missing … is the *disjoint* case" for `read32`/`write32` — that disjoint case is exactly [#131](https://github.com/cajal-technologies/talos/pull/131) (`CodeLib.RustStd.Frame`), which lifts the full read/write-disjoint family (all four 32/64 width combos) out of the old NumInteger spec. So `Mem.read32_write32_of_disjoint` / `read8_write32_of_ne` here are `Mem.read32_write32_disjoint` there; `Mem.write64_eq_write32` / `read64_toUInt32` are the kind of 32↔64 bridge that fits alongside it.

**One 32-bit twin away.** The rest of `Framing.lean` is the 32-bit mirror of the `words64` view I added in [#139](https://github.com/cajal-technologies/talos/pull/139) (`CodeLib.RustStd.MemArray`) on top of the `MemRegion` algebra in [#138](https://github.com/cajal-technologies/talos/pull/138):

| `Framing.lean` (here) | codelib (open stack) |
|---|---|
| `wordsAt` | `Mem.words64` — same `(List.range n).map …` shape, scale 4 vs 8 |
| `wordsAt_length` / `wordsAt_getElem` | `Mem.length_words64` / `getElem_words64` |
| `wordsAt_write32_of_disjoint` | `Mem.words64_write64_outside` |
| `toNat_wordAddr` | `MemRegion.slot64_base_toNat` |
| `wordsAt_congr_of_bytes` | `Mem.words64_ext` |

Your `wordsAt` is `words64` with `read32`/`4*i` in place of `read64`/`8*i`, so a `Mem.words32` twin (plus `slot32`) is a drop-in, and #139 already carries `words64_write64_extend` — the "fill one more slot" step your `copy_from_slice` phase needs the copy-analogue of.

Happy to add `words32`/`slot32` + the `memory.copy`↔view lemmas to codelib so `Framing.lean` becomes an import (matching your `wordsAt` names via a thin alias so nothing here changes) — **or** leave them in-PR if you'd rather not depend on the #68 stack mid-proof. Whichever keeps you unblocked; just let me know the preference. The allocator / `memory.grow` obligations are yours — I'm only offering the region/view plumbing.

mfornet:
I'm going to close this PR for now, as this was mostly and experiment, and we don't want this shape of proof going forward.

Issue #95, opened 2026-06-22
call_indirect is unsound — accepts a supertype, runs an ill-typed call


jasisz:
`call_indirect` checks the stored function's type against the expected one with **structural equality**. So if the table holds a function whose type is a *supertype* of the call-site type, it runs it instead of trapping. It should require a *subtype*.

```wat
(module
  (rec
    (type $super (sub (func (param i32) (result i32))))
    (type $sub   (sub $super (func (param i32) (result i32)))))
  (func $impl (type $super) (i32.add (local.get 0) (i32.const 1000)))
  (table 1 funcref)
  (elem (i32.const 0) $impl)
  (func (export "f") (result i32)
    (i32.const 7) (i32.const 0) (call_indirect (type $sub))))
```

Table holds `$impl : $super`, call site wants `$sub`. `$super` isn't a subtype of `$sub` → must trap.

- wasmtime 43 → `indirect call type mismatch`
- V8 (Node 26) → `function signature mismatch`
- **Talos → `1007`.** Runs it.

Same with GC ref results (`$impl` returns `anyref` in the table, call site wants a func returning `eqref`, `eqref <: anyref`): both engines trap, Talos returns `1`.

It's `Semantics.lean:1252` (and `:1277` for table64):

```lean
if fn.params = ty.params ∧ fn.results = ty.results then …
else .Trap st "indirect call type mismatch"
```

`$super` and `$sub` have the same params/results, so equality passes and it runs. Needs to be `fnType <: ty`, not `=`. The check works otherwise — valid direction agrees, an `i64`/`i32` mismatch traps in both — it's just the wrong relation.

Both modules validate clean. Found differential-testing against wasmtime + V8.

jasisz:
Not touching the fix — subtyping here is subtle, easy to make worse. Just flagging it.

jasisz:
Same bug in `return_call_indirect`, and the check is duplicated — `fn.params = ty.params ∧ fn.results = ty.results` appears in 4 arms (`call_indirect` i32/i64 ~1259/1284, `return_call_indirect` i32/i64 ~1114/1132). A fix needs all four or a sibling stays unsound.

```wat
(module
  (rec
    (type $super (sub (func (result i32))))
    (type $sub   (sub $super (func (result i32)))))
  (func $impl (type $super) (i32.const 1007))
  (table 1 funcref)
  (elem (i32.const 0) $impl)
  (func (export "f") (result i32)
[...]

PR #111, opened 2026-07-03
Audit-driven cleanup: examples harness, dead-code prune, doc/CI drift

El3ssar:
## Overnight audit & tidy-up: interpreter + codelib

A comprehensive, verification-driven cleanup pass over the interpreter and the
core proof/reuse layer (codelib), plus the doc and CI drift the audit surfaced.
The guiding constraint throughout: **keep the code easy to keep developing** —
no clever machinery that would make the next std function harder to add, and the
reusable-chunk story (inline-agnostic proofs landing in CodeLib) left intact and
better documented. The verifier package was explicitly out of scope.

### How this was scoped

Started from the existing 18-reviewer consolidated review (`talos-quality-review.md`),
then **re-verified every finding against current `HEAD`** with a fan-out of
per-area agents, because PRs #103/#104/#105/#110 landed after that review was
written. That re-verification changed the picture materially:

- Most U64 boilerplate findings were **already fixed by #104** (Div/Rem, Shl/Shr,
  the magic div/rem trap literals). No action needed — flagged as already-fixed.
- Several "dead code" items are actually **live via the verifier template**
  (`of_wp_entry`, the `CodeLib.UInt32` module, `TerminatesWith.mono`) — kept.
- New issues the old review predated: `CallIndirectSubtype.lean` (orphaned by
  #103) and `KvSetter.lean` (never in the codelib build) — both never checked by CI.

Only high-confidence, low-risk changes were applied. Anything requiring
speculative proof engineering with an expensive build-verify loop was
**deferred with a concrete design** (see below) rather than left half-done.

---

### What changed

**Interpreter — examples (the biggest cleanup)**
- New `Interpreter/Wasm/Examples/Harness.lean` centralises the `native_decide`
  plumbing every example re-declared: `runValues` / `runTrapMsg` / `runInvalidMsg`
  (the `Result`-projection helpers, previously copied ~20× under two spellings),
  `decodeOrDefault` (the decode-or-`default` wrapper, copied 5×), and two
  `UInt32.toNat_sub_one_*` counter-stepping lemmas (a proof block pasted 6× in
  `SimpleLoop`/`Factorial`/`EvenOddRec`). ~150 lines of duplication removed.
- Wired the two **orphaned, never-built** example files into `Examples/Basic.lean`
  so their regression `native_decide` checks actually run: `GlobalInitExpr.lean`
  (extended-const global initializers, #91) and `CallIndirectSubtype.lean`
  (the call_indirect subtype soundness fix, #95/#103).
- Deleted `Gcd.lean` and `SelectAbs.lean`: both orphaned **and rotted** (fail to
  type-check today), and both superseded — gcd by the `NumInteger` corpus proof,
  select/abs by the wired-in `SelectMin` + `IfAbs`.
- Fixed `SimpleLoop.lean`'s docstring (its stated invariant/variant disagreed
  with the proof) and a doc comment mis-attached to `emptyEnv` in `DecoderImportedGlobal`.

**CI robustness (root-cause fix)**
- The reason orphaned examples silently rot is that **CI never builds the
  interpreter library** — the main job builds only `programs/lean` (whose closure
  excludes the examples) and the smoke job builds only the `runner` exe. Changed
  the smoke job's `build-args` from `runner` to `Interpreter runner`, so the
  examples (and any future orphan) are compiled and their checks run in CI.

**Interpreter — WP layer**
- Deleted `wp.imp` (a duplicate of `wp.conseq`, no callers) and the unused
  `ImplyF` abbrev + `⇒` notation from `Wp/Defs.lean`.
- Deleted the `wp_call_of_terminates` alias (byte-identical to `wp_call_tw`);
  migrated its two `NumInteger` call sites to `wp_call_tw`, the name #105
  standardised on.
- Added a "staged for indirect-call targets" note to `wp_callIndirect_tw` (kept:
  it is the exact indirect analogue of the corpus-standard `wp_call_tw`).
- Documented the intended-public discharge API in `Spec/Defs.lean` and fixed the
  `env`-argument omission in its predicate docstrings.

**CodeLib**
- `Entry.lean`: removed the three unused tactic macros (`wasm_entry`,
  `wasm_entry_for`, `wasm_funcspec`) — zero consumers; kept `of_wp_entry`
  (verifier template) and `mono` (now heavily used post-#104/#105). Refreshed the
  module docstring to name the real corpus workhorses.
- `UInt64.lean`: deleted the unused private `ntz_le`; cleaned scratchpad
  narration comments in `ctz64_lt`; fixed a stale (pre-monorepo) docstring path.
- `UInt32.lean`, `RustStd/Option.lean`: docstrings now name the actual consumer
  (verifier template) / state the staged-ahead status, so the next dead-code
  sweep doesn't re-flag them.
- Bit-op naming consistency: renamed `and_seq`/`or_seq`/`xor_seq` →
  `bitand_seq`/`bitor_seq`/`bitxor_seq` so each `_seq` shares its `_chunk`'s Rust
  stem (updated the one consumer, `RustU64Tests/Spec.lean`).
- NEAR (kept low-touch per its lower priority): deleted the dead `setRegisterIf`
  (superseded by `checkedSetRegister?`) and nine unused `*Contract` exact-contract
  aliases; documented the `*_satisfies_rel` theorems as intentional
  build-time Env↔contract sync checks; removed the dead `finalStore`/`finalHost`
  model cluster from `KvSetter` and corrected its "proof unfinished" docstring
  (`set_spec` is a complete WP proof).

**Docs**
- `CLAUDE.md` / `AGENTS.md` / `README.md`: fixed the `programs/` → `programs/lean/`
  path and the `Programs` → `Project` package name; corrected "traps are a `none`
  result from `run`" (it returns a `Result α` with a `.Trap` constructor);
  replaced the nonexistent `@[wasm_spec]` attribute with the real `@[spec_of]` /
  `@[proves]`; added the omitted `env`/`initial` args to the `TerminatesWith` /
  `PartiallyMeets` signatures.

---

### Deliberately kept (staged-ahead API, not dead)

`wp_callIndirect_tw`, `TerminatesWith.of_run_eq` / `.toPartiallyMeets`,
`RustStd/Option.lean`, `CodeLib.UInt32`, the NEAR `*_satisfies_rel` theorems and
`resolveSpec?`/`resolveEnv?` pair — each serves the std-corpus roadmap or the
verifier template and is now annotated so it reads as intentional.

### Deferred (recommended follow-ups, with designs)

These are real but were **not** done unattended because they need speculative
proof engineering (expensive verify loops) or touch the two largest, most
heavily-depended-upon files. Each has a concrete plan in the audit notes:

1. **`Wp/Structural.lean` — factor the block/iff/loop rule skeleton.** The single
   biggest core duplication (~200 lines: the OOF-split + fuel-stabilisation + 8-way
   `exec` case tree is open-coded 3× with 5 byte-identical pass-through arms). A
   shared `wp_of_step` helper parameterised over the one-step unfold lemma
   collapses all three; public statements stay byte-identical. Medium-risk proof
   work — worth its own focused PR.
2. **Relocate the four `Mem.*` read/write frame lemmas** from `RustStd/Frame.lean`
   into the interpreter's `Mem.lean` (they are pure interpreter facts under the
   `Wasm` namespace); hoist the four pure-`Mem` disjointness lemmas currently
   buried in `NumInteger/Spec.lean` at the same time. Forces a `Semantics` rebuild,
   so batch it.
3. **Decoder dedup** (`Wat.lean`): the collect-names passes, folded-operand loops,
   `resolveTypeIdx` inlining, hex decoders, and the dead `stubImmediateCount` path
   — all behaviour-preserving but gated by the spec testsuite; plus that file's
   stale module docstring. One testsuite-validated PR.
4. **Named-case `rcases`** in `exec_loop_cons_unfold` / `exec_block_cons` /
   `exec_iff_cons` (replace fragile positional constructor lists).
5. **A single `pageSize` constant** for the `65536` literal (Semantics + Mem +
   examples + `AbsDiff` + the new `Array/FatPtrAt`); `AbsDiff` should use the
   trunk's `framePost` instead of re-spelling it.
6. **Wire `KvSetter.lean` into the codelib build** (after fixing its 6 unused-simp-arg
   lints) so the flagship NEAR theorem is CI-checked — deferred as NEAR is lower
   priority and it adds recurring `native_decide` build cost.
7. Testsuite tooling dedup + `Testsuite.lean` stale `--fuel`/docstring text
   (isolated to the testsuite exe; behaviour-preserving).

### Verification

All builds clean, **zero warnings** (CI treats `warning:` as an error):

- `cd interpreter && lake build` — 3022 jobs, incl. the whole `Examples` tree +
  both newly-wired orphans + `Harness`.
- `cd programs/lean && lake build` — 3033 jobs (interpreter core + `CodeLib` +
  `Project`); covers the renamed bit-op consumer and the `wp_call_tw` migration.
- `lake env lean CodeLib/Near/Examples/KvSetter.lean` — type-checks (this file is
  not in the codelib build; its remaining unused-simp-arg lints are pre-existing
  and untouched here — see deferred item 6).

A repo-wide grep confirmed no dangling references to any deleted/renamed symbol
(`wp.imp`, `ImplyF`, `wp_call_of_terminates`, `and_seq`/`or_seq`/`xor_seq`, the
entry macros, `ntz_le`, `setRegisterIf`, the nine NEAR `*Contract` aliases, and
the `finalStore`/`finalHost` cluster). The only same-named survivors are the
*unrelated* `HostContract HostState` contracts in the `Counter` example.

Net: **42 files, +244 / −407**. Opened as a draft for review.


El3ssar:
Split into smaller, independently-reviewable PRs (≤4 files each). Closing this in favour of those.

PR #112, opened 2026-07-04
wp: remove wp.imp and unused ImplyF

El3ssar:
wp: remove `wp.imp` and the unused `ImplyF` implication

`wp.imp` is a one-line argument-swapped wrapper around `wp.conseq` with no
callers — `wp.conseq` is the one actually used. `ImplyF` (per-state implication)
and its `⇒` notation are likewise unreferenced in code (only mentioned in prose).
Removes both; keeps `Imply`/`⇛` and `AssertionF`, which are used.


mfornet (approved):
Yay for removing lines of code, even if it is only 6 in this case.

PR #121, opened 2026-07-04
near: remove dead KvSetter model, fix docstring

El3ssar:
near: remove dead KvSetter model, fix stale docstring

`finalHost` / `finalStore` / `finalStore_satisfies_set_post` are a leftover
model cluster from an earlier proof strategy — the actual `set_spec` proof never
references them. Deletes all three and updates the module/section docstrings,
which still claimed the proof was unfinished (it is a complete WP proof).


mfornet (approved):
Agree on removing this while they are not used, but the plan is to use all of them

PR #125, opened 2026-07-04
examples: build orphaned examples, drop rotted ones

El3ssar:
examples: build the two orphaned examples, drop the rotted ones

`GlobalInitExpr.lean` (extended-const global initializers) and
`CallIndirectSubtype.lean` (the call_indirect subtype-check regression guard)
exist but were never imported by `Examples/Basic.lean`, so their `native_decide`
checks never ran. This wires both into `Basic.lean`. It also deletes `Gcd.lean`
and `SelectAbs.lean`, which are likewise unimported *and* no longer type-check,
and are already covered elsewhere (gcd by the `NumInteger` proof, select/abs by
`SelectMin` + `IfAbs`).


mfornet:
I'm closing this as this examples should be tested now

PR #131, opened 2026-07-05
codelib: lift NumInteger's memory framing lemmas into RustStd.Frame (#68 phase 1)

theebayuser:
## Overview

Phase 1 of #68. Every memory-based program proof needs the same read-after-write algebra — "a read at `B` is unaffected by a write at disjoint `A`" — but the six disjointness lemmas lived as private copies inside the NumInteger spec, so every new program (next up: `SwapElementsSpec` from #110) would re-derive them from scratch. This PR moves them into `CodeLib.RustStd.Frame`, next to the existing read-after-write-`same` lemmas, and points NumInteger at the shared copies.

## What changed

**`codelib/CodeLib/RustStd/Frame.lean`**
- New byte-level footprint lemmas: `Mem.write64_bytes_of_disjoint`, `Mem.write32_bytes_of_disjoint`.
- New word-level framing lemmas, all four width combinations: `Mem.read64_write64_disjoint`, `Mem.read64_write32_disjoint`, `Mem.read32_write32_disjoint`, `Mem.read32_write64_disjoint`.
- Module docstring extended to document the split: unconditional `_same`/`_pages` lemmas stay global `@[simp]`; the `_disjoint` family is deliberately **not** `@[simp]` (each rewrite carries a disjointness side condition), so proofs name them explicitly in `simp only` sets / `rw` and discharge the side goal with `decide`/`omega` — exactly how NumInteger already uses them.

**`programs/lean/Project/NumInteger/Spec.lean`**
- The "Memory framing lemmas" section (~85 lines) is deleted; the six use sites now reference the `Mem.`-prefixed CodeLib names. Proof scripts are otherwise untouched.

Statements are moved **verbatim** (only namespaced under `Mem.`): the `.toNat` disjunction shape is load-bearing, because existing call sites discharge the side conditions with `(by decide)` and `simp (config := { decide := true })` on concrete frame addresses.

## Scope notes for review

- `Mem.read32_write32_disjoint` / `Mem.read32_write64_disjoint` currently have no call site (NumInteger only consumes the `read64_*` pair; the `read32_*` pair completes the width matrix and existed in-tree already). Happy to drop them from this PR if you'd rather they land with their first consumer.
- `MemRegion` and `write_write_commute` from #68 are intentionally **not** here — I'd like to drive them out against a real consumer, the `SwapElementsSpec` proof, as a follow-up PR.

## Testing

- `cd programs/lean && lake build` — full interpreter → codelib → programs chain, all `@[proves]` theorems re-elaborate.
- No interpreter/decoder changes; `testsuite_report.txt` untouched.

## Disclosure

Developed with AI assistance (Claude Code); I've reviewed and understand every line and am happy to answer questions or address feedback.

### DIFF
--- a/codelib/CodeLib/RustStd/Frame.lean
+++ b/codelib/CodeLib/RustStd/Frame.lean
@@ -16,14 +16,23 @@ form that does not care about the concrete spill address:
 * `Mem.read{32,64}_write{32,64}_same` — read-after-write at the same
   address returns the value (no bound on the address needed).
 * `Mem.write{32,64}_pages` — a store leaves the page count unchanged.
+* `Mem.read{32,64}_write{32,64}_disjoint` — a read is unaffected by a
+  write to a disjoint byte range, in every width combination, built on
+  the byte-level `Mem.write{32,64}_bytes_of_disjoint`. Proofs that juggle
+  several frame slots (argument spills vs. scratch vs. result) need these
+  to carry a value across the unrelated stores in between.
 
 The in-bounds (no-trap) obligation also needs `sp - 16` not to underflow;
 that is `UInt32.toNat_sub_of_le`, which Lean core already provides
 (`Init.Data.UInt.Lemmas`), so it is used directly rather than re-proved here.
 
-The four `Mem.*` lemmas are intentionally global `@[simp]` — confluent,
-terminating rewrites used corpus-wide. Op-specific lemmas (e.g. `popcnt`
-bounds) are added here only once a real proof first consumes them.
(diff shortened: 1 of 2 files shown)


mfornet:
I'm closing this PR as I understand it was superseded by #137, correct me if I'm wrong

theebayuser:
That's right — #137 was stacked on this branch, so its squash-merge (`1c41a12`) carried the `Frame.lean` changes and the NumInteger refactor with it. Nothing was lost. Thanks!

PR #132, opened 2026-07-06
Float Programs

Isabelle9999:
## Overview
Follow-up to #62 (the 76 float wp lemmas). The lemmas gave `wp_run` 
the ability to step through every float instruction with an `execOne` 
case; this PR adds the programs that exercise them: real Rust → wasm32 (opt-level 0) → decoded Program.lean → proven TerminatesWith spec. Each program is an equivalence check between a naive and an optimized implementation of the same operation, specifically targeting inputs where IEEE-754 edge cases (NaN, rounding tie-breaks, saturation) could cause them to disagree.

Three of the four planned programs are fully proven (zero sorry). The second (array min/max) ships its compiled Program.lean and Rust source, but its spec is intentionally deferred: at opt-level 0 the compiler never emits the f32.min/f32.max instructions the program was designed to exercise. Details in the Program 2 section below.

## What changed
**Program 1 — `float_trunc`: safe float-to-int conversion** (proven, zero sorry)
- naive: NaN check via `f32.ne(x,x)` + range check + trunc; optimized: 
  `i32.trunc_sat_f32_s`
- Range check uses strict `f32.lt`/`f32.gt` because 2^31−1 is not 
  exactly representable in f32 (avoids the boundary off-by-one — the 
  Ariane 5 failure class, Lions et al. 1996)
- Exercises: conversions + comparisons

**Program 3 — `float_round`: rounding behavior** (proven, zero sorry)
- naive: `frac = x − trunc(x)`, threshold 0.5, ceil/floor; optimized: 
  `f32.nearest` (round-half-to-even, IEEE 754-2019 §4.3.1)
- The two intentionally disagree on exact half-integers (different 
  rounding modes); the spec proves termination with a defined result 
  for all inputs
- Exercises 11 instructions: f32.nearest/trunc/ceil/floor/sub/ge/le/
  eq/load/store/const — the unary + arithmetic lemma families

**Program 4 — `float_reinterpret`: precision and bit reinterpretation** (proven, zero sorry)
- Check A: `f32.abs` vs clear-sign-bit via 
  `i32.reinterpret_f32 / i32.and 0x7FFFFFFF / f32.reinterpret_i32` 
  (reinterpret preserves all bits; abs is defined as sign-bit clear)
- Check B: `f32.abs` vs promote-abs-demote (`f64.promote_f32` / 
  `f64.abs` / `f32.demote_f64` — no precision loss since magnitude 
  unchanged)
- Check C: `f32.copysign` vs bit manipulation
- Exercises: all four reinterpret directions + demote/promote + 
  remaining unary

**Program 2 — `float_minmax`: array min/max** (scaffolded; spec deferred, documented)
- Design constraint discovered during implementation: at opt-level 0, 
  Rust's `f32::min`/`f32::max` compile to comparison branches 
  (`f32.lt`/`f32.gt` + select), NOT the `f32.min`/`f32.max` 
  instructions — LLVM only emits the min/max instructions at opt-level 
  ≥ 1. Since the project standard is non-optimized code, the compiled 
  program cannot exercise the min/max lemmas as designed.
- The program as compiled still exercises f32.load/store, comparisons, 
  and loop control flow; Program.lean + Rust source are included, spec 
  intentionally deferred pending a decision (see design notes).
- The `f32.min`/`f32.max` wp lemmas remain correct, forward-looking 
  infrastructure for optimized or hand-written wasm.

**Float-opacity resolution — `CodeLib/IEEE32/Exec.lean`**
Three proof obligations in Program 1 reduce to facts about `Float32` 
that Lean's kernel cannot evaluate (Float is opaque). Resolved via a 
bit-level IEEE-754 binary32 model on `UInt32` (extracted from 
TorchLean): pure bitvector definitions for isNaN/comparisons/saturating truncation, closed by `bv_decide` (SAT — seconds, vs `native_decide`'s infeasible 2^32 enumeration), bridged to the runtime `Float32` by five explicit axioms stating the Lean runtime implements IEEE-754 binary32. The axioms are the standard trust 
boundary for this problem (same approach as TorchLean's FP32 
soundness layer).

**On FloatSpec compatibility** : I attempted to make  FloatSpec (Beneficial-AI-Foundation) compatible instead of introducing axiom. but there was a toolchain mismatch (FloatSpec pins Lean v4.29.0 while Talos is on v4.31.0) and FloatSpec's float type (`PrimFloat`) is currently backed by the real numbers with `is_nan` hardcoded to `false` ( it's a mathematical placeholder, not a model of actual IEEE-754 bit-level behavior, and it has no connection to Lean's runtime `Float32`). Since this need bit-level behavior (NaN, saturation), FloatSpec cannot express them.

The IEEE32Exec axioms are designed so that each axiom states one fact of the form "Lean's runtime Float32 operation X matches the IEEE-754 bit-level definition." 

Generated with Claude Code


mfornet:
Thanks for this examples, I'll merge this for now, as this are great examples on how to start working with floats.

mfornet (approved):

PR #136, opened 2026-07-10
programs: add encode/decode string↔bytes crates (wasm + emitted Lean)

El3ssar:
## What

Two new verification-target crates under `programs/rust/`, each following the standard crate template (pure fn in `lib.rs` + `extern "C"` wrapper in `exports.rs`):

- **`encode`** — `fn encode(s: &str) -> Vec<u8>`: a string to its UTF-8 bytes (`s.as_bytes().to_vec()`).
- **`decode`** — `fn decode(bytes: &[u8]) -> String`: bytes back to the original string (`String::from_utf8(...).unwrap()`). Inverse of `encode`; the round-trip is identity since a `&str` is already UTF-8.

## Scope

This is the first slice: **compiled wasm + transpiled Lean only, no proofs yet.**

- `Program.lean` is emitted by `verifier emit` from the compiled `.wat` and type-checks (`lake env lean`).
- `Spec.lean` is the emit scaffold stub (`:= True`) — to be filled in.
- The modules are intentionally **not** wired into `programs/lean/Project.lean`, so the default proof build is unaffected.

Specs + proofs will follow in subsequent commits on this branch.

## How it was produced

`just verifier-add {encode,decode}` → `just verifier-build` → `just verifier-emit`.


mfornet:
I'm gonna close this for now, as I think this is a very large effort that will need to be repeated once iris-lean integration lands.

PR #137, opened 2026-07-11
programs: prove SwapElementsSpec (swap_elements total correctness)

theebayuser:
## Overview

Discharges `SwapElementsSpec`, the statement-only spec registered by #110 ("No proof is attempted here"). The exported `swap_elements(ptr, len, i, j)` provably terminates and swaps the `u64` elements at indices `i` and `j` in place, leaving every other in-bounds element unchanged — total correctness via `TerminatesWith`, no fuel in the statement.

**Stacked on #131** (consumes the `CodeLib.RustStd.Frame` disjointness family; the first commit here is that PR). Uses only the repo's existing WP layer — no new dependencies.

## Proof structure

One `TerminatesWith` lemma per function in the opt-0 call graph, composed at each `call` with `wp_call_tw`:

- `func2_swap` — the exchange leaf (`tmp := arr[i]; arr[i] := arr[j]; arr[j] := tmp` through the scratch slot at `global0 − 16 + 8`). Stated as an explicit final-memory equation so callers frame reads themselves.
- `func1_swap` — drives the three nested bounds-check `block`s; `i, j < len` makes both `panic`/`unreachable` branches dead, i.e. the Rust bounds checks are proven unreachable.
- `func0_swap`, `func3_writes` — the forwarder and the fat-pointer materialisation.
- `func4_swap` — the export: threads frame setup (`global 0 := 1048560`), the fat-pointer round-trip, teardown, and decodes the three-write trace into the read-based postcondition (with the degenerate `i = j` self-swap case handled).
- `elemAddr_toNat` / `elemAddr_disjoint` — no-wrap address algebra: distinct in-bounds indices give 8-byte-disjoint addresses.

## The registered spec is strengthened — deliberately

Two preconditions are added to the #110 sketch, and the module docstring documents why each is **necessary for soundness**, not convenience:

1. **`st.globals.globals[0]? = some (.i32 1048576)`** (shadow-stack pointer at its initial value). The callee frames are derived from `global 0`; unpinned, they can alias the array and the postcondition is falsified.
2. **`st.mem.pages ≤ 65536`** (the wasm32 memory ceiling; the module declares `pagesMin = 17`). Without it the original four hypotheses are satisfiable with wrap-around: e.g. `ptr = 2³² − 8`, `len = 131071` puts element `k = 131070` at address `1048552` — exactly the scratch slot — so the frame clause of the *original* statement is provably false.

Both mirror the shadow-stack pin already assumed by the `total_variation` proof.

## Verification

- `lake build` green across the whole `programs` package; zero `sorry`, zero lint warnings.
- `#print axioms swap_elements_correct` → `propext, Classical.choice, Quot.sound` plus the two LRAT-checked `bv_decide` reflection axioms inherited from `Frame.lean`. No `sorryAx`.
- `Program.lean` untouched (canonical module from #110); single-file diff to `Spec.lean`.

Refs #110, #68.

---

*Disclosure per CONTRIBUTING: AI tooling (Claude Code) was used to write and check this proof; I own the change and am accountable for every line.*

### DIFF
--- a/codelib/CodeLib/RustStd/Frame.lean
+++ b/codelib/CodeLib/RustStd/Frame.lean
@@ -16,14 +16,23 @@ form that does not care about the concrete spill address:
 * `Mem.read{32,64}_write{32,64}_same` — read-after-write at the same
   address returns the value (no bound on the address needed).
 * `Mem.write{32,64}_pages` — a store leaves the page count unchanged.
+* `Mem.read{32,64}_write{32,64}_disjoint` — a read is unaffected by a
+  write to a disjoint byte range, in every width combination, built on
+  the byte-level `Mem.write{32,64}_bytes_of_disjoint`. Proofs that juggle
+  several frame slots (argument spills vs. scratch vs. result) need these
+  to carry a value across the unrelated stores in between.
 
 The in-bounds (no-trap) obligation also needs `sp - 16` not to underflow;
 that is `UInt32.toNat_sub_of_le`, which Lean core already provides
 (`Init.Data.UInt.Lemmas`), so it is used directly rather than re-proved here.
 
-The four `Mem.*` lemmas are intentionally global `@[simp]` — confluent,
-terminating rewrites used corpus-wide. Op-specific lemmas (e.g. `popcnt`
-bounds) are added here only once a real proof first consumes them.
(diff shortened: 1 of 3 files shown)


mfornet:
Welcome to the project, and thanks for your contribution 🙌

mfornet (approved):

PR #142, opened 2026-07-13
num_integer: opt-level 3 build + proved opt0/opt3 gcd equivalence

El3ssar:
## What

Adds a second build of the existing `num_integer::gcd_u64` at **opt-level 3**,
a **reusable program-equivalence predicate** in CodeLib, and a **proved
equivalence** between the two builds. **`gcd_opt_equiv` is proved — no `sorry`**
(`#print axioms` is clean). CI builds all of it.

## The two builds

`num_integer_opt3` is a byte-for-byte copy of the `num_integer` source,
compiled at opt-level 3 via a per-package profile override
(`[profile.release.package.num_integer_opt3]`); the workspace default stays 0.

| | opt-level 0 (`num_integer`) | opt-level 3 (`num_integer_opt3`) |
|---|---|---|
| functions | 3 (`func0`/`func1`/`func2`) | **1**, fully inlined |
| linear memory | spills operands + Stein scratch (shadow stack) | **never touched** |
| `global 0` (stack pointer) | mutated, then restored | **never touched** |
| `gcd_u64` export | func **2** | func **0** |

Both modules declare the same mutable state (16-page memory, globals
`[1048576×3]`, one 1×1 funcref table), so `mod0.initialStore = mod3.initialStore`.

Empirically cross-checked with node over 1225 `u64` pairs (incl. `(0,0)`,
powers of two, `u64::MAX`): opt0 ≡ opt3 ≡ reference gcd, 0 mismatches.

## Reusable predicate — `CodeLib/Equivalence.lean`

`Wasm.ObservationallyEquiv env m₁ id₁ m₂ id₂ initial args`: two entry points,
run from the same initial store on the same args, reach exactly the same
observable outcomes — **returned values + host state** (`Store.host`), success
and trap symmetric — **linear memory deliberately not observed**.

Ships fully-proven (no `sorry`): `ObservationallyEquiv.of_common_outcome` (the
discharge rule, via `TerminatesWith.outcome_unique`) plus `refl`/`symm`/`trans`.

## The equivalence — `Project/NumIntegerOpt3/Equivalence.lean`

`GcdOptEquiv` instantiates the predicate at the two builds from the canonical
store; `gcd_opt_equiv` proves it by reducing to a common outcome both builds
reach — each `gcd_u64` terminates with the same value (the gcd) and the same
host state.

- opt0 side reuses the existing `Project.NumInteger.Spec.gcd_u64_correct`.
- opt3 side is the new `NumIntegerOpt3/Spec.lean`: **`mod3_gcd`** proves the
  register-only build computes the gcd and leaves the store untouched. Its core
  is **`inner_wp`**, a `wp_loop_cons` invariant for Stein's subtract-and-halve
  loop reusing the CodeLib `UInt64` Stein lemmas (`stein_step_x/y`,
  `shr_ctz_*`, `recombine_loop`); the surrounding structure (odd-part
  reduction, zero-checks, recombine) is driven by two small local macros.

The `mod0.initialStore` starting point is **load-bearing**: on a pathological
store the opt0 build can trap (shadow-stack pointer / pages) where the
memory-free opt3 build still returns, so the two are *not* equivalent for an
arbitrary initial state.

## Notes for review

- No `sorry` anywhere; `#print axioms gcd_opt_equiv` = the standard axioms plus
  the interpreter's two `bv_decide` axioms (inherited from the opt0 memory
  lemmas). CodeLib addition is `sorry`-free.
- All three modules (`Spec`, `Equivalence`, and the CodeLib predicate) are wired
  into the build, so CI checks them.
- The auto-scaffolded `Spec.lean` placeholder was replaced by the real
  `mod3_gcd`; the equivalence lives in `Equivalence.lean`.


programs/lean/Project/NumIntegerOpt3/Equivalence.lean
```diff
@@ -0,0 +1,119 @@
...
+intended, general shape of the observation. -/
+def GcdOptEquiv : Prop :=
+  ∀ (env : HostEnv Unit) (initial : Store Unit) (a b : UInt64),
+    initial = mod0.initialStore →
+    ∀ (result : List Value) (hostFinal : Unit),
+      TerminatesWith env mod0 entry0 initial [.i64 a, .i64 b]
+          (fun st vs => vs = result ∧ st.host = hostFinal)
+        ↔
+      TerminatesWith env mod3 entry3 initial [.i64 a, .i64 b]
+          (fun st vs => vs = result ∧ st.host = hostFinal)
```
mfornet: bring this out of this folder, and make it reusable over multiple programs.

El3ssar: Done, the equivalence function is added to codelib and the spec became a oneliner with the parametrized ObservationallyEquiv function called on both modules and functions.

mfornet (approved):

PR #143, opened 2026-07-13
Iprop module-linking

Isabelle9999:
## Overview

module linking :: function specs carry iProp ownership (pointsTo_u64, separating conjunction ∗), so the frame rule guarantees that callers' private memory is untouched across function calls automatically without manual disjointness proofs.

This is the standard Iris-Wasm pattern (PLDI'23, §2.2) adapted for Talos's big-step fuel semantics.

## Why we needed `wp_wasm_iProp`

Talos's existing `wp_wasm` takes `Q : Store Unit → List Value → Prop` , which is a Prop-level postcondition. This works for intra-module proofs (swap, merge_sort) where Prop composition suffices. But for module 
linking, function specs need iProp postconditions (pointsTo_u64, ∗) so the frame rule operates at the iProp level where ∗ is native.

The solution: a parallel `wp_wasm_iProp` that takes `Φ : Store Unit → List Value → IProp WasmHeapGF` as postcondition. Both WPs coexist, existing proofs use `wp_wasm` unchanged, module linking uses `wp_wasm_iProp`. A bridge theorem connects them:

```lean
wp_wasm_iProp_pure : wp_wasm Q = wp_wasm_iProp (fun st' vs => ⌜Q st' vs⌝)
```

This is Talos-specific: Iris-Wasm (small-step) has wp_bind built into Iris generically. Our big-step fuel semantics requires explicit fuel composition (`wp_wasm_iProp_call`), which is the standard adaptation every Iris instantiation makes for its language (HeapLang has `wp_rec`, Iris-Wasm has `wp_invoke_native`, we have `wp_wasm_iProp_call`).

This parallel architecture means each per-construct rule (block, call, loop) needs both Prop and iProp versions, which cost more maintenance cost.

Generated with Claude Code


mfornet:
Merge this on top of latest branch of #133 to keep them in sync

Isabelle9999:
> Merge this on top of latest branch of #133 to keep them in sync
I will rebase after the iprop fix, because funcSatisfies needs iProp postconditions for the frame rule

codelib/CodeLib/SepLogic/AllocSpec.lean
```diff
@@ -0,0 +1,22 @@
+import Interpreter.Wasm
+
+namespace Wasm.SepLogic
+
+open Wasm
+
+axiom dlmalloc_alloc_spec
```
mfornet, line 7: @El3ssar this might be enough for you for the moment for the encode/decode experiment

@Isabelle9999 please open an issue tracking this axiom: the goal is to convert it to a theorem

mfornet (approved):

codelib/CodeLib/SepLogic/AllocSpec.lean
Isabelle9999: #158

Isabelle9999:
rebased onto #156 and got rid of conflicts

Isabelle9999:
closing to adapt to small-step

PR #151, opened 2026-07-15
codelib+programs: memory-aware program equivalence, proved on swap_elements opt0/opt3

theebayuser:
## Overview

`ObservationallyEquiv` (#142) observes the returned values and `Store.host`, and deliberately omits linear memory. Its docstring left a TODO:

> Programs whose *result* lives in memory need a stronger observation than this one; **add it when such a proof arrives**.

`swap_elements` is exactly that program. It returns `[]` and communicates only by mutating the caller's array, and its `Store.host` is `Unit` — so the existing notion degenerates on it to bare *co-termination*: true, but it says nothing whatsoever about the swap. This PR adds the stronger observation, and the proof that motivates it: **`swap_elements` at opt-level 0 vs opt-level 3, related through the array itself.**

Everything is proved — no `sorry`, `#print axioms` clean.

## The two builds

Byte-for-byte the same Rust source (`arr.swap(i, j)` on `&mut [u64]`), compiled at two optimisation levels via a per-package profile override, mirroring the `num_integer_opt3` pattern from #142.

| | opt-level 0 (`swap_elements`) | opt-level 3 (`swap_elements_opt3`) |
|---|---|---|
| export | func **4** | func **0** |
| structure | 16-byte shadow-stack frame, fat pointer through memory, four-deep call chain | fully **inlined** into one function |
| `global 0` (stack pointer) | mutated, then restored | **never touched** |
| the exchange | via a **scratch slot** at `1048552` | two `i64.load`s / `i64.store`s through an `i64` local |
| preconditions needed | shadow-stack pin + `1048576 ≤ ptr` | **neither** |

## Why the observation must include memory — but not *all* of memory

The observation is

```lean
fun st => (st.host, st.mem.words64 ptr len.toNat)
```

the host state plus **the caller's array, viewed as a `List UInt64`** (`Mem.words64`, #139).

It is deliberately weaker than "the final memories are equal", and it has to be — the two final memories are **not** the same function. opt0 additionally writes the exchanged value into the scratch slot at `[1048552, 1048560)`, which opt3 never touches. Observing the array *region* rather than all of memory is exactly what separates the caller-visible result from caller-invisible scratch traffic. That is the whole point of the exercise, and it is what `Mem.words64` is for.

### `codelib/CodeLib/Equivalence.lean` — the observation becomes a parameter

Rather than clone the predicate per observation, `ObservationallyEquivOn … (obs : Store α → β)` states the core **once**, with `outcome_unique_on` / `of_common_outcome` / `refl` / `symm` / `trans` against it. (`run` is a function of fuel, so the final store — hence any observation of it — is pinned; the existing proof generalises verbatim with `congrArg obs`.)

`ObservationallyEquiv` becomes the instance at `Store.host`. **Every statement is unchanged**; its proofs are now one-liners delegating to the generic versions. `Project.NumIntegerOpt3.Equivalence` builds **untouched** — I verified `git diff` on that file is empty. This follows the `write_write_comm_of_footprints` precedent from #147: a new observation now costs an instance, not another copy of the proofs.

### `codelib/CodeLib/RustStd/MemArray.lean` — `Mem.words64_swap`

The view-level counterpart of a per-element swap postcondition: `m'`'s array view is `m`'s with positions `i` and `j` exchanged. Stated at the slot addresses, so it needs **no no-wrap hypothesis**, and `i = j` is allowed (both sides degenerate). Consumer: this PR's equivalence, via one shared bridge used by *both* sides.

### `programs/` — the opt3 build and the proof

- `swap_elements_opt3` crate + emitted `Program.lean`.
- `SwapElementsOpt3.Spec.func0_swap` — total correctness of the inlined export.
- `SwapElementsOpt3.Equivalence.swap_opt_equiv` — the equivalence. The opt0 side **reuses the merged `swap_elements_correct`** (#137) as-is; both sides route through `words64_of_swap_post`, so they land on the same observation and `of_common_outcome` applies.

Per the opt3 twin convention (`NumIntegerOpt3/Spec.lean`), the opt3 build carries no `@[spec_of]` — that would duplicate the opt0 registration for the same Rust function.

## Verification

- `lake build` green in `codelib/` and `programs/lean/`; zero warnings.
- `#print axioms swap_opt_equiv` → `[propext, Classical.choice, Quot.sound]` plus the two pre-existing `bv_decide` reflection axioms from `read{32,64}_write{32,64}_same`. No `sorryAx`.
- `verifier check --no-prove` reports every existing `Program.lean` up to date — no drift.
- `Project.NumIntegerOpt3` (the existing consumer of the generalised predicate) rebuilds with no changes to that file.

Refs #142, #68. Builds on #137 / #139.

---

*Disclosure per CONTRIBUTING: AI tooling (Claude Code) was used to write and check these changes.*


theebayuser:
Review pointer: ~7,070 of the +7,554 lines are `SwapElementsOpt3/Program.lean` — `verifier emit` output, checked by the "Programs up to date" CI gate. The hand-written surface is four files: `codelib/CodeLib/Equivalence.lean` (observation-generic core; every existing statement byte-unchanged, `NumIntegerOpt3/Equivalence.lean` untouched), `RustStd/MemArray.lean` (`words64_swap`), `SwapElementsOpt3/Spec.lean` (opt3 total correctness), and `SwapElementsOpt3/Equivalence.lean`. @El3ssar — this instantiates the extension point your docstring described ("Programs whose result lives in memory need a stronger observation than this one; add it when such a proof arrives"), at `fun st => (st.host, st.mem.words64 ptr len.toNat)`. Happy to split the codelib generalisation into its own PR if preferred.

mfornet:
I'm ok merging this now, but keep in mind that we are doing an effort to rewrite all proofs dealing with storage (memory, etc...) using iris-lean which is a nice separation logic framework.

theebayuser:
Sounds good. The equivalence layer should carry over as-is — `of_common_outcome` only consumes `TerminatesWith` facts and doesn't care how they were proven, so once the iris-lean bridge lands, adequacy can discharge the same statements. Happy to help with the rewrite wherever useful.

mfornet:
@theebayuser please address the conflicts, I think they were introduced after merging #133

theebayuser:
Done — merged main in. The only conflict was the import line in `programs/lean/Project.lean`, where #133's `SwapSepLogic` import landed next to the two `SwapElementsOpt3` ones; kept both. Green locally on codelib and programs (including the new SepLogic files), and `#print axioms swap_opt_equiv` is unchanged.

Issue #108, opened 2026-06-25
call_indirect runs an ill-typed call across reordered recursion groups


jasisz:
Talos computes recursion-group type identity **equi-recursively** (by unrolling / bisimulation) rather than **iso-recursively** (by positional rec-group structure). Two recursion groups that hold the same mutually-recursive types in a different **member order** are DISTINCT types under the spec's canonicalization, so a `call_indirect` against one type on a function of the other must trap. Talos treats them as equal and **executes the ill-typed indirect call**.

## Minimal repro
```wat
(module
  ;; G1 = [A ; B] :  A = func(ref B) -> i32,  B = struct(ref A)
  (rec
    (type $A1 (func (param (ref null $B1)) (result i32)))
    (type $B1 (struct (field (ref null $A1)))))
  ;; G2 = [B ; A] :  same two types, reordered
  (rec
    (type $B2 (struct (field (ref null $A2))))
    (type $A2 (func (param (ref null $B2)) (result i32))))
  (func $f (type $A1) i32.const 77)
  (table 1 funcref)
  (elem (i32.const 0) $f)
  (func (export "go") (result i32)
    (call_indirect (type $A2) (ref.null $B2) (i32.const 0))))   ;; $f : $A1, called as $A2
```
| engine | result |
|--------|--------|
| wasmtime 46.0.0 | TRAP — `indirect call type mismatch` |
| WasmEdge 0.17 | TRAP — `Expected ref type_index[2], Got ref type_index[1]` |
| V8 (Node) | TRAP |
| **Talos** | **`77`** — runs the call |

## Why this is a bug
Wasm GC is **iso-recursive**. Canonicalization rolls each recursion group, substituting a group's internal type references with **positional (de-Bruijn) recursive indices**, so the canonical form depends on **member order**. `[A;B]` (A at position 0, refers to B at position 1) and the reordered `[B;A]` (A now at position 1, refers to B at position 0) therefore give A **different canonical types**, even though their unrolled infinite trees coincide — that distinction is exactly the difference between iso-recursive and equi-recursive typing. So `$A1 ≠ $A2`, and `call_indirect (type $A2)` on a `$f : $A1` must trap. WasmEdge's own diagnostic spells out the structural difference (`type_index[2]` vs `type_index[1]`).

## Controls (rule out a malformed probe)
[...]

Issue #107, opened 2026-06-24
Use differential fuzzing tool to keep parity with trusted interpreted


mfornet:
> This is the differential tool I've been using — cleaned it up a bit and put it out as its own thing, separate from Talos (it drives any runtime through a CLI shim and diffs against V8 / wasmtime / the spec interpreter): https://github.com/jasisz/miscast
> 
> One heads-up so the output doesn't surprise you: it also flags a few f64 literal mismatches — every one I traced is Lean stdlib rounding (`Float.ofScientific` / `Float.ofNat`), i.e. upstream, not Talos's own logic.
>  

 _Originally posted by @jasisz in [#95](https://github.com/cajal-technologies/talos/issues/95#issuecomment-4787516274)_

Kayvan-Zahiri:
Happy to take this on as the in-repo *integration* of @jasisz's [miscast](https://github.com/jasisz/miscast) (crediting it, not rebuilding): a `just differential` recipe that drives the `runner` exe as the SUT, a small seed corpus, and a CI gate that fails only on a real soundness divergence — with the known f64 `Float.ofScientific` / `Float.ofNat` rounding mismatches @jasisz flagged filtered out as upstream noise. Differential oracles / eval harnesses are what I worked on at Outlier, so this is in my lane.

@jasisz — would you rather miscast stay an external dep I pin/vendor, or have a thin driver mirrored in-repo? Happy to coordinate on Telegram first. I'll keep it to one focused PR (adapter + seeds, then CI separately).

Kayvan-Zahiri:
Quick prototype update — wired @jasisz's miscast to the talos `runner` locally via a thin adapter (maps miscast's `{wat} {wasm} {export}` to `runner <file> <method> [args]`). Two useful findings:

- **V8 works as an oracle out of the box** via `node` — no wasmtime install needed for a first CI lane.
- `--mode recgroup --sut custom` surfaces **300 soundness reproducers, and they all share one signature** — spec/V8 `TRAP`, talos returns a value for an ill-typed `call_indirect` across reordered rec groups — i.e. they minimize to the same root cause as **#108**. Direct check, no miscast in the loop: `runner module.wat go` returns `932` where V8 traps. So the harness already pins a real, known bug at scale; the 300→1 clustering (and filtering the f64 `Float.ofScientific`/`Float.ofNat` rounding noise you flagged) is exactly the triage I'd wire in.

[...]

PR #152, opened 2026-07-16
Add `just differential`: differential-test the runner against V8 via miscast

Kayvan-Zahiri:
Part of #107 — the runnable harness. The CI gate follows as a second PR, as discussed in the thread.

Adds `just differential`, which keeps the runner in parity with a trusted oracle (V8) by driving it as the custom system-under-test for [@jasisz](https://github.com/jasisz)'s [miscast](https://github.com/jasisz/miscast), and flags divergences — above all **soundness**, where the runner accepts or runs a module every oracle rejects or traps on.

## Design

The runner plugs into miscast with **no adapter**: miscast invokes a custom SUT as `CUSTOM_CMD='<cmd> {wat} {export}'` and reads a value from stdout / a trap from a `trap:` line, which is exactly the runner's CLI and output contract. Out-of-fuel (exit 2, empty stdout) reads as unsupported, so it is never a false divergence.

miscast is pinned as an external dependency (a rev in `scripts/differential.sh`), not vendored. The first run clones it into `.differential-cache/` (gitignored); set `MISCAST_DIR=` to use a local checkout.

## What it turns up

- **#108 looks fixed.** The seed `differential/seeds/recgroup_callindirect.wat` (the minimal reordered-rec-group `call_indirect`) now traps with `indirect call type mismatch`, agreeing with V8. It is kept as a regression guard.
- **A full sweep (`just differential --mode all`) surfaces 27 candidate soundness divergences** of the #95 / #108 class — the runner accepts and runs spec-invalid modules that both V8 and `wasm-tools validate` reject — clustering into ~6 families: subtyping/variance, const-init typing, result/arity, array/table reftype ops, exception-handler typing, and `ref.test`/`br_on_cast`. I hand-checked three; all accept and run. Each is reproducible from the harness.

  Happy to fold these into a known-divergence baseline for the CI gate, or file the sharpest as individual issues — whichever you prefer.

## Contents

- `scripts/differential.sh` — build the runner, pin miscast, run against V8
- `differential/seeds/recgroup_callindirect.wat` — minimal #108 reproducer, now a regression guard
- `differential/README.md`, a `just differential` recipe, a `CONTRIBUTING.md` section, and a `.gitignore` entry

## Requirements

`wasm-tools` and `node` >= 22 (the V8 / WasmGC oracle).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ verifier-report/
 
 extracted/
 out/
+
+# differential-testing scratch (pinned miscast clone + run output)
+.differential-cache/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,6 +56,17 @@ The recipe requires the pinned `wasm-tools` version (see `WASM_TOOLS_VERSION` in
 
 Bumping `WASM_TOOLS_VERSION` or the `vendor/testsuite` submodule will also change the report — those bumps and the regenerated report belong in the same commit.
 
+## Differential testing
+
+Beyond the spec testsuite, `just differential` keeps the runner in parity with a trusted engine (V8) by running the same modules on both and flagging divergences — above all **soundness** divergences, where the runner accepts or runs a module the oracle rejects or traps on.
+
+```bash
+just differential                          # recgroup soundness mode, V8 oracle
+just differential --mode recgroup -n 300   # reproduce the full #108 cluster
+```
+
+It drives the runner as the custom system-under-test for [miscast](https://github.com/jasisz/miscast) (pinned as an external dependency). Requires `wasm-tools` and `node` ≥ 22. See [`differential/README.md`](differential/README.md) for details.
+
 ## Contributing code
 
 Pull requests are welcome. A few guidelines:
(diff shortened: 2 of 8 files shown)


Kayvan-Zahiri:
Thanks for the fixes. The arg-count wording is a good catch now rather than later: `_VALERR` greps for `"mismatch"`/`"expected"`, so if a CLI arg-count error looked like a validator rejection, we'd end up recording false REJECT agreement on invalid modules. I had checked the value/trap path against miscast's regexes but not the error paths, so the canaries are a good addition.

Fresh sweep on the current branch (`--mode all -n 50`):

`SOUNDNESS=27 VALUE=2 OVER_TRAP=6 OVER_REJECT=13 SUT_UNSUP=5 AGREE=145`

`SOUNDNESS` and `VALUE` are unchanged from the pre-fix run, so they aren't artifacts of the float fixes. The two `VALUE` cases also aren't float-formatting related: `castalgebra-funcsub[wrong-variance]` and `[wrong-result-covariance]` return `1` where V8 returns `0`, so the runner is accepting a function subtype it shouldn't.

The fixes did improve overall parity, though: `SUT_UNSUP` dropped from `8 → 5` and `AGREE` increased from `142 → 145`.

Happy to turn this into the CI-gate baseline. Thanks.

mfornet (approved):

PR #175, opened 2026-08-11
codelib: total WP for the u32 copy loop

theebayuser:
## Summary

The copy loop in `RustStd/MemCopyLoop.lean` only had a Löb-based partial rule
(`copyWords_smallStep_wp`), so it proved the copy correct but not that it
finishes. This adds the total version alongside it, closed with the `n - i`
variant. Three pieces:

- `twp_shl` in `SepLogic/SmallStepTotalLifting.lean`, the one pure rule the loop
  needed that was not there yet (the address arithmetic is
  `localGet, const 2, shl, add`).
- `twp_loop_wf_family` in a new `SepLogic/SmallStepTotalLoop.lean`. This is the
  total twin of `wp_loop_löb_family`, which already lives in
  `SmallStepLifting.lean:860` next to `loopBodyExpr`. With no later to guard the
  recursive occurrence, the body instead gets the recursive hypothesis at
  strictly smaller measures.
- Six `copyWords_*_twp`, mirroring the existing partial chain one for one, over
  a `CopyWordsLoopState` index carrying the counter, the copied prefix, and the
  two unprocessed tails.

No new axioms: the total chain has the same footprint as the partial
`copyWords_smallStep_wp` it mirrors.

## Notes

`MergeSort/TotalProof.lean:22` and `Quicksort/Laws.lean:852` each have a
`Unit`-specialised twin of this loop rule. I left both alone rather than touch
files that just landed. Happy to switch them over to the shared one if you want
that, either here or as a follow-up.

The rule stops where the partial rule stops, at the Iris WP, since it stays open
in the framed resource and its continuation. Reaching a `TerminatesWith` needs a
closed config the way `Examples/Gcd.lean` and `Quicksort/TotalProof.lean` do, so
that would be a separate example rather than part of this rule.

### DIFF
--- a/codelib/CodeLib/RustStd/MemCopyLoop.lean
+++ b/codelib/CodeLib/RustStd/MemCopyLoop.lean
@@ -1,5 +1,6 @@
 import CodeLib.RustStd.MemArray
 import CodeLib.SepLogic.SmallStepLifting
+import CodeLib.SepLogic.SmallStepTotalLoop
 
 /-!
 # A verified copy loop over a `u32` array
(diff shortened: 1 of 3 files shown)


mfornet (approved):

theebayuser:
@mfornet  Thanks for the merge. The termination ledger table in IirisMigration.md is still TBD, and with gcd, mergesort, and quicksort having total forms now, it looks fillable. I did a pass over the corpus for the intended-measure column; most rows are just recording which specs are partial-only right now. Do you want a PR for the table?

mfornet:
At this point I think we can remove the IrisMigration document, and continue working from here.

PR #181, opened 2026-08-15
Add `@[rust_ref]` and `@[equiv_of]` to CodeLib.Attrs

El3ssar:
`codelib/CodeLib/Attrs.lean` gains two attributes beside `spec_of` and `proves`:

```lean
syntax (name := rust_ref) "rust_ref" str   : attr  -- @[rust_ref "crate::fn"]
syntax (name := equiv_of) "equiv_of" ident : attr  -- @[equiv_of modelName]
```

`@[rust_ref]` marks a `def` as a hand-written Lean model of the named Rust function; the target is quoted for the same reason `spec_of`'s is, a Rust path being no Lean identifier. `@[equiv_of]` marks a theorem as the equivalence that model owes to the compiled function — it names a Lean declaration, so it takes a bare identifier and registers the way `proves` does. Both are `.afterCompilation` no-ops like the existing pair. Neither enforces the obligation; together they make owed and discharged visible in the source rather than inferred from naming.

An attribute is only available to files that import its definition, so `Attrs.lean` cannot apply its own registrations and nothing there breaks if a grammar drifts. `codelib/CodeLib/Attrs/Check.lean` is that application site: one tagged `def`, one tagged `theorem`, nothing claimed about wasm. It is a `lean_lib` root of its own rather than part of the `CodeLib` umbrella, so `import CodeLib` does not pull the two placeholders in. Happy to fold it into the umbrella, move it under `Examples/`, or drop it.

The extractor is untouched. It reads `spec_of` and `proves` and ignores other attribute heads, so the new tags reach no artifact field and raise no diagnostic, and a `def` carrying only `@[rust_ref]` is not a `FormalSpec`. Recording them would be a schema change; `EXTRACT.md` now states that gap rather than leaving it to be discovered.

### DIFF
--- a/codelib/CodeLib/Attrs.lean
+++ b/codelib/CodeLib/Attrs.lean
@@ -1,14 +1,15 @@
 import Lean
 
 /-!
-# `@[spec_of]` and `@[proves]` — load-bearing project attributes
+# The load-bearing project attributes
 
-These two attributes are the link between Lean specifications and the code
-they describe, and between proofs and the specifications they discharge.
-They carry the metadata that `verifier extract` reads off the source. The
-runtime behavior here is intentionally minimal — the attribute machinery
-only needs to *exist* so the source typechecks; semantics live in the
-extractor.
+These four attributes are the link between Lean specifications and the code
+they describe, between proofs and the specifications they discharge, and
+between a hand-written Lean model and the compiled function it stands in
+for. They carry the metadata that `verifier extract` reads off the source.
+The runtime behavior here is intentionally minimal — the attribute
+machinery only needs to *exist* so the source typechecks; semantics live in
+the extractor.
 
 ## `@[spec_of <kind> "qualified::name"]`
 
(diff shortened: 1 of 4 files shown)


El3ssar:
Closing: these two declarations belong in the consuming project, not in CodeLib.

`spec_of` and `proves` are CodeLib's because CodeLib's own extractor reads them. `rust_ref` and `equiv_of` describe a convention imposed by our workspace scaffolder and read by our own checker (`equivalence`), never by `verifier extract` — so putting them here would ask CodeLib to carry schema for a downstream tool that does not consume them, and would make every consumer of CodeLib import two attributes that mean nothing to it.

They now ship in the scaffolded workspace itself, as an `[ours]` `lean/Project/Attrs.lean` imported by the workspace's seam file, in the same registration shape used here. A scaffolded workspace elaborates `@[rust_ref "crate::fn"]` and `@[equiv_of modelName]` with no change to this repository. The branch is left in place; nothing else here is touched.

PR #186, opened 2026-08-18
Build mergesort wasm size-first and regenerate Program.lean

mfornet:
## What

Compile the `mergesort` crate (and its only dependency, `talos-stdio`, used by no other crate) for the shortest wasm binary instead of the workspace's unoptimized default, via per-package `[profile.release.package.*]` overrides — the same precedent as `num_integer_opt3` / `swap_elements_opt3`. This is also the answer to "how do we configure how programs are compiled": the workspace `Cargo.toml` is the single source of truth, and every CI job picks it up automatically because they all run the same cargo workspace build.

Along the way, this PR makes the verifier's wasm build **byte-identical across machines**, and regenerates the corpus with that pipeline.

## Size measurements

Fully stripped `build/mergesort/program.wasm`, rustc 1.95.0, canonical verifier build:

| opt-level | size |
|---|---|
| 0 (old default) | 55,865 B |
| `"z"` | 40,807 B |
| `"s"` | 40,423 B |
| **3 (shipped)** | **40,416 B** |

Counterintuitively, plain `opt-level = 3` beats both nominal size-first levels on this crate; the numbers are recorded next to the override.

## Reproducibility fixes

#185 made ubuntu CI the canonical wasm build on the premise that cross-platform byte-reproducibility was unattainable. Two concrete divergences were the real causes, both fixed here:

1. **rust-src path leak** — hosts with the `rust-src` component embed panic-location paths from the sysroot src tree; hosts without embed `/rustc/<commit-hash>/…` paths of a different length, shifting the data section at higher opt levels. `reproducibleRustflags` now remaps the src tree onto the `/rustc/<hash>` prefix.
2. **Symbol-hash leak in the name section** — mangled names carry cargo's per-unit hash, seeded by machine-specific inputs, and the emitted `Program.lean` embeds the printed wat verbatim. `wasm-tools strip --all` drops the name section (code/data unaffected).

Additionally, wasm-tools is pinned to 1.251.0 (same as the justfile) in both workflows, since its wat printing feeds the embedded drift check. #185's build-once artifact handoff in CI is kept, but local regeneration on any machine now reproduces CI byte-for-byte — no artifact download needed.

## Verification

- Full `lake build` over the regenerated corpus: green, 3,549 jobs, 0 sorries.
- `Spec.lean` for mergesort needed no changes — `module_imports` still closes by `rfl`, so the StdIO host contract and `MergesortSpec` carry over to the opt-3 binary.
- The freshness check on earlier revisions of this branch already demonstrated a macOS-emitted corpus passing ubuntu CI byte-for-byte.


mfornet:
Pushed a second commit addressing the failing **Verifier Freshness** check (and the proof CI, which turned out to be red on main too):

**How program compilation is configured.** Compile flags per program live in `programs/rust/Cargo.toml` as `[profile.release.package.<crate>]` overrides — the verifier pipeline and both CI jobs run the same cargo workspace build, so they pick these up automatically. `mergesort` (and `talos-stdio`) build at `opt-level = 3`, which measured smallest. This is now documented in the freshness workflow.

**Why freshness was failing** (three independent causes):
1. *Pre-existing on main since 5ea231c:* the emitter format changed (block result lists, `Function.typeIdx`, embedded wat drift-check) but the corpus was never regenerated — CI re-emits from scratch, so every crate drifted. All 14 `Program.lean` are now regenerated, and the spec files are migrated (function-record literals carry the emitted `typeIdx`; SepLogic specs take the new host-state parameter, `WasmSmallStepGS hlc Unit` etc.).
2. *rust-src path leak:* with the `rust-src` component installed locally, rustc embeds panic locations from the sysroot src tree; CI (no rust-src) embeds `/rustc/<hash>/…` paths of a different length, shifting the wasm data section at opt-level 3. `reproducibleRustflags` now remaps the src tree onto the `/rustc/<hash>` prefix.
3. *wasm-tools drift:* the emitted `Program.lean` embeds the printed wat verbatim, so the printer version matters; CI floated on `@latest` (1.256.0) vs. the pinned 1.251.0. The workflow now pins 1.251.0, matching the justfile.

Full `lake build` over the migrated corpus is green locally — 3,549 jobs, 0 sorries.

mfornet:
Rebased on main now that #185 has landed. The branch is a single commit with the remaining delta:

- **mergesort at `opt-level = 3`** (plus `talos-stdio`), regenerated `Program.lean` — 40,416 B fully stripped vs 55,865 B at the opt-0 default; the nominal size-first levels measure slightly larger ("s" 40,423 B, "z" 40,807 B).
- **Machine-independent wasm builds**, superseding #185's premise that only ubuntu CI bytes are canonical: the `/rustc/<commit-hash>` remap for hosts with `rust-src`, `wasm-tools strip --all` (the name section leaked machine-seeded symbol hashes into the embedded wat), and wasm-tools pinned to 1.251.0 in both workflows. #185's artifact handoff to the macOS proofs job is kept as a build-once optimization — but local regeneration on any machine now reproduces CI byte-for-byte, no artifact download needed.
- All 14 `Program.lean` re-emitted with the reproducible pipeline; #185's spec migration is untouched. Full `lake build`: green, 0 sorries.

PR #183, opened 2026-08-15
M8  total correctness

Isabelle9999:
Total correctness for 6 remaining programs via TWP, completing M8.

## Overview

Adds TWP proofs and `SmallStep.TerminatesWith` for FloatTrunc, FloatRound, FloatReinterpret, SwapElements, SwapElementsOpt3, and NumInteger. Each program's `@[spec_of]` Prop is redefined from legacy big-step `TerminatesWith env` to `SmallStep.TerminatesWith`. Both cross-optimization equivalence proofs (NumIntegerOpt3, SwapElementsOpt3) ported to `SmallStep.ObservationallyEquivOn`.

Merge sort adequacy closure (`mergesort_partiallyMeets` + `mergesort_terminatesWith`) completes M7 item 8.

## New shared infrastructure

 `wasm_smallStep_heap_globals_runtime_store_terminates` in `SmallStepAdequacy.lean`:  TWP adequacy for programs that own heap memory and globals. The existing `wasm_smallStep_heap_store_terminates` discards globals; all Float programs and SwapElements need the stack-pointer global.

## Retained legacy

`SwapElements/Spec.lean` keeps legacy theorems because `swap_spec_sep` in `SwapSepLogic.lean` has `SwapElementsSpec` as its theorem type. New small-step specs are registered and live. Cleanup is M9 scope.

FloatMinmax (stub) and Near layer (blocked on `wp_callHost`) documented in `M8_MIGRATION_LEDGER.md`.

## Per-PR Checklist

1. Milestone: M7 item 8 + M8.
2. Oracle: Legacy `TerminatesWith env` theorems.
3. Step/step?: No new transitions.
4. Soundness: Not applicable.
5. Nondeterminism: No.
6. Boundary cases: Covered by existing WP proofs and regression tests.
7. Framing: Same ownership as existing WP proofs.
8. Ported: FloatTrunc (3), FloatRound (7), FloatReinterpret (12), SwapElements (5), NumInteger (5), both equivalence proofs. Full ledger in `M8_MIGRATION_LEDGER.md`.
9. Build: `lake build --wfail` in interpreter, codelib, programs/lean. Zero errors, warnings, sorry.
10. Atomicity/ownership: No changes.
11. Open items: SwapElements legacy cleanup, FloatMinmax stub.

PR #192, opened 2026-08-20
Fix main: adapt small-step TWP adequacy to the new iris-lean

El3ssar:
Main is red. #183 and #188 each passed alone but conflict once merged: #188 updated iris-lean so `twp.to_wp` is now a **wand**, and #183's proof passes it to `Entails.trans`, which needs an **entailment**.

One real change, in `SmallStepAdequacy.lean`:

```lean
-exact (htwp .hasLC).trans twp.to_wp
+exact (htwp .hasLC).trans (wand_entails twp.to_wp)
```

The other 6 lines follow from it. The old proof left a universe free, which gave `wasm_smallStep_heap_*` an accidental universe parameter that never appeared in their statements; call sites pinned it with `.{0}`. `wand_entails` determines that universe, so the parameter is gone and the `.{0}` annotations have to go with it.

Builds green: codelib 3530, programs 3564.


El3ssar:
Superseded by #190, which lands the same fix (same 7 files, same `wand_entails` change and `.{0}` removals). Closing in favour of it.

PR #195, opened 2026-08-24
Rule out normal completion when unreachable is pending

El3ssar:
**If the outermost frame is about to run `unreachable`, the program can never finish normally.**

That is the whole content. It lets you prove a module does *not* meet a specification, which is currently hard to do.

## What is added

One predicate and three theorems in `interpreter/Interpreter/Wasm/SmallStep.lean`, next to `safe_of_steps_to_done`:

```lean
def Config.CannotComplete (config : Config α) : Prop

theorem Config.CannotComplete.of_step :          -- preserved by every Step
    config.CannotComplete → Step config kind next → next.CannotComplete

theorem Steps.invariant :                        -- lift any such predicate over Steps
    (∀ …, P config → Step config kind next → P next) → Steps config trace final → P config → P final

theorem Config.CannotComplete.not_steps_to_done : -- the conclusion
    config.CannotComplete → Steps config trace ⟨.done values, store⟩ → False
```

Only the outermost frame matters, because `finish` and `returnFromFunction` are
the only constructors producing `.done` and both need an empty call stack.
Everything nearer the top is unconstrained — the callee may return, trap, throw
or diverge.

## Why

The existing theorems here — `steps_done_deterministic`, `steps_comparable`,
`safe_of_steps_to_done` and the rest — all compare **two executions you already
have**. That is the right tool for reconciling two proofs and the wrong one for
refuting a specification: to show a module fails `TerminatesWith` you would
need a competing terminal trace, and if the callee may diverge there is none.

`CannotComplete` is decided by the shape of the configuration alone, so it needs
no execution, and it survives arbitrarily much unanalysed execution in between.

## Two side conditions, both necessary

Each is falsified in the PR rather than asserted, so the hypotheses cannot be
weakened by mistake later:

- **`unreachable` must be the *next* instruction**, not merely present. A branch
  can jump over it into the function's normal exit — `terminatesWith_of_branch_past_unreachable`.
- **The outermost frame's control stack must be handler-free.** `unwindExceptionCall`
  discards the caller's continuation, so a `tryTable` there can catch and complete
  normally with the `unreachable` never executed — `terminatesWith_of_catch_past_unreachable`.

## Checked

`#print axioms` is `[propext, Classical.choice, Quot.sound]` for every theorem;
no `sorry`, no `native_decide`, default heartbeats. Builds against `main` at
`3d742fb`, and `codelib` with `--wfail` is green. The block adds ~20s to a file
that already takes ~11.5 minutes.

### DIFF
--- a/interpreter/Interpreter/Wasm/SmallStep.lean
+++ b/interpreter/Interpreter/Wasm/SmallStep.lean
@@ -7600,6 +7600,397 @@ theorem safe_of_steps_to_done
   · obtain ⟨suffix, hdone⟩ := towardDone
     exact checked_ok_of_steps_to_done hdone
 
+/-! ### Frames with no normal exit left
+
+Every theorem above compares two executions that somebody has already
+exhibited. Nothing above says anything about a configuration no execution has
+been exhibited from — yet that is exactly what is needed to *refute* a
+specification: a callee may diverge, so there is no trapping trace to compare
+against, and determinism has nothing to work with.
+
+`Config.CannotComplete` is the invariant that fills the gap. It holds of a
+configuration whose outermost frame can no longer return normally, it is
+preserved by `Step`, and it is false of `⟨.done _, _⟩` — so by `Steps.invariant`
+no execution from such a configuration ever completes normally, whatever
+happens in between.
+
+The outermost frame is the one that decides: `finish` and `returnFromFunction`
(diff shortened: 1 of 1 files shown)


mfornet:
I'll close this for now, since this feature is not being used, we can bring it back in the future when we start using it.

PR #215, opened 2026-08-28
codelib: drop the NEAR import matcher in favour of the shared one and add a contract tactic

El3ssar:
- `Near.importMatches` was byte-identical to `ImportDecl.matches` in the interpreter's `Host/Registry.lean`; removed in favour of the shared one.
- ~21 `*_satisfies_rel` theorems all had the same proof script; introduced a `near_contract` tactic macro.
- Corrected a stale docstring on `nearEnv_satisfies_canonical` that asks for a variant `HostRegistry.envFor_satisfies` already provides.

Deliberately does **not** attempt the full `HostRegistry` migration — `resolveImports?` returns `none` on an unknown import while `entryFor` is total and substitutes a trapping stub, so that is a behavioural change needing its own review.

Verified: green (3599 jobs).

### DIFF
--- a/codelib/CodeLib/Near/Env.lean
+++ b/codelib/CodeLib/Near/Env.lean
@@ -1043,14 +1043,11 @@ def nearImports : List ImportDecl := nearHostFns.map Prod.fst
 `nearImports`. -/
 def nearEnv : HostEnv NearState := { funcs := nearHostFns.map Prod.snd }
 
-/-- Import declarations match by module, name, params, and results. -/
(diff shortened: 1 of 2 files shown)


El3ssar:
Closing on review. Three unrelated changes in 79 lines, and the `near_contract` macro refactors proofs that a separate NEAR cleanup deletes outright. The one clearly-good part — dropping `Near.importMatches` in favour of the identical `ImportDecl.matches` — should ride with that NEAR work instead.

PR #218, opened 2026-08-28
codelib: add wp_liftDetStep and use it for the globals rule family

El3ssar:
`SmallStepLifting.lean` repeats a primStep-inversion block ~107 times (~2100 lines); its 6-line core is byte-identical in ~100 of them, with only the `step_deterministic` argument varying. `wp_pureStep` is already the store-preserving degenerate case.

Adds `wp_liftDetStep`, the stateful generalisation (next expression + store transformer + a store predicate, with `==∗` premises in the style of the existing `wp_callHost`), and converts the globals family to it.

Scoped to globals on purpose — tables and memory are separate PRs. Statements unchanged. Verified: green (3599 jobs).

### DIFF
--- a/codelib/CodeLib/SepLogic/SmallStepLifting.lean
+++ b/codelib/CodeLib/SepLogic/SmallStepLifting.lean
@@ -61,6 +61,73 @@ theorem wp_pureStep
   · iexact Hwp
   · itrivial
 
+/-- Generic lifting rule for a deterministic Wasm step that may also update the
+store. `wp_pureStep` is the degenerate store-preserving case; the stateful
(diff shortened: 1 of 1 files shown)


El3ssar:
Closing on review. This converts only 2 of ~103 call sites in `SmallStepLifting.lean` and makes the file net longer — the combinator is right, but it needs to land with the store/local/table families so it actually pays for the concept it introduces. Worth reopening as one PR that takes `wp_localSet`, `wp_localTee`, `wp_tableSet`, `wp_dataDrop`, `wp_elemDrop` and the `wp_store*` family together.

PR #224, opened 2026-08-28
codelib: small mechanical dedup in RustStd and the UInt lemmas

El3ssar:
Small mechanical cleanups in `RustStd/`:

- `absDiffHeap_pointsTo` re-derived in ~62 lines what `store64Heap_pointsTo` already concludes, even though `absDiffHeap` is *defined* as `store64Heap ∅ …`. Now pipes through it (~25 lines), mirroring the 32-bit sibling.
- Hoisted a twice-copied `heq` block; replaced two inlined empty-heap proofs with the existing lemmas.
- `fatPtrArithmetic` took a whole `FatPtrAt` but used only `h.noWrap`; weakened to `hroom`, with `fatPtrArithmetic_of` preserving the original for existing call sites.
- Renamed genuinely-unused `hxodd`/`hyodd` to `_hxodd`/`_hyodd` rather than dropping them (~10 call sites pass them by name).
- Dropped the private `_aux_toNat_UInt64_ofNat` for core's `UInt64.toNat_ofNat'`. Kept `_aux_toNat_pos` — core does not supersede it.

**−141/+69.** Green (3599 jobs).


mfornet:
@copilot resolve the merge conflicts in this pull request

PR #229, opened 2026-09-02
programs: add the rust_vec crate with a Vec model and wire codec

theebayuser:
## Summary

- rust_vec: six exports over the talos-stdio byte stream. Every export reads one borsh value and writes one borsh value (`from_slice` / `to_vec`, `default-features = false`).
- `vec_push` is new. `vec_pop` now writes `(Option<u32>, Vec<u32>)` and `vec_get` writes `Option<u32>`. `vec_len`, `vec_contains`, `vec_sum32` keep their shape. Tell me if the last two should go.
- codelib: `CodeLib.RustStd.Borsh` names the borsh layouts over `Vec.serialize` (`u32`, `bool`, `option`, `vec`, `vec?`, round trip `vec?_vec`). A tuple is `++`.
- Lean: six partial contracts (`WritesOrOOM`, the `talos.oom` disjunct as in Mergesort), stated with those layouts. Proved: the round trip, the three module facts, `contract_names_start` (every export name starts a call), and five readers on well-formed input (`push`, `get`, `pop`, `pop` on empty, `sum32`).
- This PR does not prove the contracts themselves. That proof needs the allocator carried through the borsh decoder, and it is a separate PR.

## Verification

- cargo fmt --check; cargo test -p rust_vec (7 tests); cargo clippy --target wasm32-unknown-unknown --all-targets -- -D warnings
- just verifier-check --force-emit --no-prove twice, byte-identical output
- lake build --wfail in codelib and lake build Project --wfail; just runner-smoke
- The proofs use no native evaluation axiom. The module facts close by `decide +kernel`.
- module: 13,034 bytes, 93 functions (25,107 and 119 with borsh's default features)

## Open question

`verifier extract` knows `rust-exported`, `rust-internal`, and `lean`. The extractor warns about `rust-exported-partial` (7 uses across two crates) and drops it, so these specs show no Rust link in the report. Should the extractor learn the tag?


programs/rust/rust_vec/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    write(&(data.len() as u32).to_le_bytes());
+}
+
+/// Write the last byte, or nothing when the input is empty.
+pub fn vec_pop() {
+    let mut data = read_all();
+    if let Some(byte) = data.pop() {
+        write(&[byte]);
+    }
+}
```
mfornet: I'd make the interface similar to a functional programming. For example:

```
vec = read_vec()
element = vec.pop()
write(element, vec) // write the element and the remaining vec
```

In this case, element can be an option, which requires some custom serialization, I recommend using [borsh](https://borsh.io/) (https://crates.io/crates/borsh) for serializing and deserializing everything.

mfornet (changes requested):
Let's include in this program

- vector length (done)
- vector push
- vector pop
- vector get

programs/rust/rust_vec/src/lib.rs
theebayuser: sounds good

PR #236, opened 2026-09-11
Improve float_minmax specification and verification

UjjwalAshiwal:
Upgraded the existing float_minmax Lean specification by replacing the True placeholder with actual check_min/check_max specifications.

Added IEEE-754 edge-case proofs for NaN, ±0, and ±Inf, along with a concrete Wasm execution check for check_min.


UjjwalAshiwal (commented):
upgraded

